PR #16, opened 2016-06-07
Salt worker refactor

marioncobretti:
This change is [Reviewable](https://reviewable.io/reviews/plus3it/watchmaker/16)


watchmaker/managers/base.py
```diff
@@ -191,40 +211,49 @@ def download_file(self, url, filename, sourceiss3bucket=False):
...
-                  '    filename = {1}'.format(url, filename))
+                          '    url      = {0}\n'
+                          '    filename = {1}'.format(url, filename))
+
+    def call_process(self, cmd):
+        if not isinstance(cmd, list):
+            exceptionhandler('Command is not a list.\n{0}'.format(str(cmd)))
+        rsp = subprocess.call(cmd)
+
+        if rsp != 0:
```
lorengordon: What do you think of just returning `rsp` and checking the return code where the process is called? We will have more context about what is failing and can provide more informative error messages that way.

marioncobretti: `def call_process(self, cmd):
     if not isinstance(cmd, list):
          exceptionhandler('Command is not a list.\n{0}'.format(str(cmd)))
     process = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
     out, err = process.communicate()
     if process.returncode != 0:
          exceptionhandler('Command execution failed.\n{0}\nError Message:{1}'.format(cmd, err)
     else:
          logging.info(out)`

I wrote this in the browser... so if there is any off syntax, ill catch it before commit.  This changes the flow a little.  It will either dump if the command execution failed and carry the command used and return the shell error in the log and output in the exceptionhandler, or it will log any output that may have come from the command.

lorengordon: Ok. I guess if call_process() throws the exception and we want to provide a more context-aware error message, we can put those particular calls in a try/except block.

marioncobretti: I dont think that will work.  Since our exceptionhandler uses sys.exit(), i dont think the exception block with execute since the process will die before a return.  If you want to be able to handle a failed call, we will need to use a flag to redirect the exception.  (i think this was your idea when we were talking about it earlier)  Something like this:

```
	def call_process(self, cmd, returnexception=False):
        if not isinstance(cmd, list):
            exceptionhandler('Command is not a list.\n{0}'.format(str(cmd)))
        process = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
        out, err = process.communicate()
        if process.returncode != 0 and returnexception:
            return subprocess.CalledProcessError, err
        elif process.returncode !=0:
            exceptionhandler('Command execution failed.\n{0}\nError Message:{1}'.format(cmd, err)
        else:
            logging.info(out)
```

lorengordon: Ahh, ok. Do you want to rework call_process() or the exceptionhandler in another PR? We can open an issue to track the enhancement and mark this discussion resolved in reviewable.

lorengordon:
Reviewed 4 of 4 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/plus3it/watchmaker/16)*

PR #31, opened 2016-06-28
Worker Refactor and Windows Provisioning

marioncobretti:
Pulling in branches SaltWorkerRefactor and Windows_Provisioning.

---
This change is [Reviewable](https://reviewable.io/reviews/plus3it/watchmaker/31)


KevinPlus3:
`from src.watchmaker import Prepare` in scripts/watchmaker.py seems weird - is it because one is assuming that watchmaker is not installed when using this script within the context of this directory?

I would assume that if watchmaker has been installed as a module into Python and then I was trying to use watchmaker.py somewhere other than the code directory, then `from src.watchmaker import Prepare` would break. I sure would not expect `src.watchmaker` to be the standard way of importing the module ... we would have to fix that if that is the outcome of using src folder.

KevinPlus3:
Not sure if we want authors to be tracked in these files?

`__author__ = 'marion'` in bootstrappers/__init__.py.

KevinPlus3:
Reviewed 19 of 32 files at r4, 1 of 4 files at r6, 2 of 3 files at r7, 6 of 7 files at r8.
Review status: 20 of 22 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/plus3it/watchmaker/31)*

KevinPlus3:
Reviewed 1 of 32 files at r4, 1 of 1 files at r9.
Review status: 20 of 22 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/plus3it/watchmaker/31)*

marioncobretti:
Good catches.  The src.watchmaker is wrong its a factor of pycharm and a mistake.  The __init__.py with my author is the same.  I forgot to turn those off on my PyCharm on my laptop (on by default).  Thanks.

KevinPlus3:
Reviewed 3 of 32 files at r4, 3 of 3 files at r10.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/plus3it/watchmaker/31)*

Issue #40, opened 2016-08-09
PowerShell bootstrapper updates


lorengordon:
Couple notes:

- [x] Rename `Version` param to `PythonVersion`
- [x] Move [python package installs](https://github.com/plus3it/watchmaker/blob/windows_installing_salt/src/watchmaker/static/shell/bootstrappers/watchmaker-bootstrap.ps1#L96-L109) to a dedicated function. `Get-PythonPackages()` or something.

lorengordon:
Fixed in #45

PR #45, opened 2016-08-10
Powershell Bootstrapper Update

KevinPlus3:
Updates to the PowerShell script to reflect Issue #40.

---
This change is [Reviewable](https://reviewable.io/reviews/plus3it/watchmaker/45)

### DIFF
--- a/src/watchmaker/static/shell/bootstrappers/watchmaker-bootstrap.ps1
+++ b/src/watchmaker/static/shell/bootstrappers/watchmaker-bootstrap.ps1
@@ -94,11 +94,6 @@ function Get-Python {
   [Environment]::SetEnvironmentVariable("Path", "${env:Path};C:\Python27\;C:\Python27\Scripts\", "Machine")
   Log "Added Python directories to environment variable, PATH."
   Reset_Env_Variable
-  
-  # Upgrade pip.
-  # pip install --upgrade pip : Do not use this as it throws an access denied error.
-  # easy_install -U pip
-  # Log "Upgraded pip using easy_intall."
 }
 
 function Get-Python-Packages {
--- a/src/watchmaker/static/shell/bootstrappers/watchmaker-bootstrap.ps1
+++ b/src/watchmaker/static/shell/bootstrappers/watchmaker-bootstrap.ps1
@@ -26,6 +26,8 @@ Param(
   ,
   [String]$RootCertUrl
 )
+# Python modules to install/upgrade with Python.
+[System.Collections.ArrayList]$Packages = @("pip")
 
 # Location to save files.
 $SaveDir = ${env:Temp}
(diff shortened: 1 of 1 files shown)


src/watchmaker/static/shell/bootstrappers/watchmaker-bootstrap.ps1
```diff
@@ -57,6 +57,23 @@ function Install-Python-MSI {
   Start-Process "msiexec.exe" -ArgumentList ${Arguments} -Wait
 }
 
+function Install-Python-Packages {
```
lorengordon: Since these installs are so similar, I kind of had in mind a generic function for installing python packages, rather than hard code the dependencies... The function could take a param `[String[]]$Packages` and install them all... What do you think?

src/watchmaker/static/shell/bootstrappers/watchmaker-bootstrap.ps1
```diff
@@ -87,26 +104,13 @@ function Get-Python {
...
-  # Install Watchmaker and Python dependencies for WatchMaker.
-  if( ${WatchmakerUrl} -eq "" ) {
-    Log "The url to a Watchmaker Python egg is empty, unable to install Watchmaker."
-  } else {
-    easy_install ${WatchmakerUrl}
-    Log "Installed Plus3's Watchmaker and dependencies for Python."
-    python -c 'import watchmaker'
-    Log "Tested that Watchmaker installed."
-  }
+  Install-Python-Packages
```
lorengordon: How about moving this call out of the `Get-Python` function, and calling it right after the call later to `Get-Python`?

src/watchmaker/static/shell/bootstrappers/watchmaker-bootstrap.ps1
KevinPlus3: Makes sense to me - I had wondered about setting it up that way so anyone can add more python modules as they see fit.

KevinPlus3: I moved the pip upgrade back into Get-Python because it is actually an installation of a new package but rather an upgrade of a package that already comes with Python's MSI installation.

src/watchmaker/static/shell/bootstrappers/watchmaker-bootstrap.ps1
```diff
@@ -87,26 +89,21 @@ function Get-Python {
...
   Download-File $PythonUrl $SavePath
   Install-Python-MSI ${SavePath}
-  Log "Installed Python ${Version}."
+  Log "Installed Python ${PythonVersion}."
 
   [Environment]::SetEnvironmentVariable("Path", "${env:Path};C:\Python27\;C:\Python27\Scripts\", "Machine")
   Log "Added Python directories to environment variable, PATH."
   Reset_Env_Variable
-
+
```
lorengordon: A bit of whitespace lint there....

lorengordon:
Reviewed 1 of 1 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/plus3it/watchmaker/45)*

Issue #47, opened 2016-08-11
salt state cli argument addition.


MarionTheBull:
This will override the config.yaml file.

--no-salt-state

MarionTheBull:
Merged with PR #48

Issue #26, opened 2016-06-22
Update mkdocs Documentation with Watchmaker Details and Usages


KevinPlus3:
Documentation about Watchmaker (its installation, usage, details, etc.) has now been migrated to the mkdocs format.  It needs to be filled out and completed.

MarionTheBull:
Merged and resolved with PR #48

Issue #2, opened 2016-06-01
Support installing config.yaml from a URL


lorengordon:

marioncobretti:
Suggestion:
--config-file
Accepts local, or http(s) url
Validates that it exists
Validates that it is yaml
If Validation fails, should it just WARN and continue with default?  or should it halt and return that it doesnt exist or validate?

lorengordon:
i vote for halt. if the user passes a value and it's bad, we should tell them it's bad so they can fix it, rather than maybe continuing to reference defaults they were trying to override.

MarionTheBull:
Merged and resolved with PR #48

PR #48, opened 2016-08-12
Issues #47 #34 #26 #2 --saltstates argument added, changed boto3/botocore imports, updated config.yaml to accept URLs as well as path to file, Usage Documentation updates

MarionTheBull:
This change is [Reviewable](https://reviewable.io/reviews/plus3it/watchmaker/48)


lorengordon:
Woah, pasting in the binary zip file is _not_ how we're doing that! This branch cannot be merged until those binaries are excised from the commit history. Not just deleted in a new commit; they must be removed from the commit history entirely. Otherwise the size of the git project starts to explode because git cannot diff binaries very well.

setup.py
```diff
@@ -2,20 +2,24 @@
...
 from __future__ import absolute_import, print_function
 
+import os
+
 from setuptools import find_packages, setup
 
 setup(
     name='watchmaker',
-    version='0.0.1',
+    version='0.1',
```
lorengordon: 0.0.1 was correct for now.

src/watchmaker/__init__.py
```diff
@@ -2,8 +2,10 @@
 import logging
 import os
 import platform
+import shutil
 import subprocess
-
+import urllib
```
lorengordon: should this be `from six.moves import urllib` for py2/py3 compatibility?

src/watchmaker/__init__.py
```diff
@@ -61,14 +79,21 @@ def __init__(
...
+            self.logger.error('{0} does not exist'.format(arguments.log_path))
         else:
             self.logger.warning('Stream logger is not enabled!')
 
         self.logger.info('Parameters:  {0}'.format(self.kwargs))
         self.logger.info('System Type: {0}'.format(self.system))
 
+    def _validate_url(self, url):
+
+        if not validators.url(url):
```
lorengordon: Can this just be `return validators.url(url)`? Though the validators docs say this returns an object on failure, wouldn't an object still evaluate to `True`?

https://validators.readthedocs.io/en/latest/#module-validators.url

setup.py
```diff
@@ -2,20 +2,24 @@
...
     author='Plus3IT Maintainers of Watchmaker',
     author_email='projects@plus3it.com',
     url='https://github.com/plus3it/watchmaker',
     packages=find_packages('src'),
     package_dir={'': 'src'},
     include_package_data=True,
+    scripts=[os.path.join('scripts', 'Watchmaker.py')],
     install_requires=[
         "botocore",
         "boto3",
```
lorengordon: Didn't one of the commits remove `botocore` and `boto3` as actual dependencies?

setup.py
```diff
@@ -2,20 +2,24 @@
...
     name='watchmaker',
-    version='0.0.1',
+    version='0.1',
     author='Plus3IT Maintainers of Watchmaker',
     author_email='projects@plus3it.com',
     url='https://github.com/plus3it/watchmaker',
     packages=find_packages('src'),
     package_dir={'': 'src'},
     include_package_data=True,
+    scripts=[os.path.join('scripts', 'Watchmaker.py')],
```
lorengordon: Surely, there is some way to do this without the capital W filename.

MarionTheBull: Surely there is but I would like to resolve the creation with this, and the open a separate issue for the work around to get the PR in.  Ill do some comparison with the entrypoint option.  Ill create a smaller issue for it.

src/watchmaker/__init__.py
MarionTheBull: Current:
```
import validators

def valid(url):
    if not validators.url(url):
        return False
    else:
        return True        

valid('https://www.google.com')
True

valid('https://www.google.c')
False
```

Suggested:
```
def valid(url):
    return validators.url(url)
   
valid('https://www.google.com')
True

valid('https://www.google.c')
ValidationFailure(func=url, args={'public': False, 'value': 'https://www.google.c'})
```

I think the current approach is correct so that we dont have use a try to validate.

lorengordon:
Reviewed 1 of 5 files at r1, 2 of 21 files at r2, 5 of 7 files at r3, 2 of 2 files at r6.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/plus3it/watchmaker/48)*

PR #52, opened 2016-08-24
Deploys dev builds to testpypi and tagged releases to pypi

lorengordon:
*   Updates .travis.yml with the configuration data to deploy to
    both pypi and testpypi
*   Adds the script, `ci/travis_set_build.py`, which will update the
    `__version__` string with the Travis build number

The idea is to use the `skip` param of the new script in conjunction
with the `$TRAVIS_TAG` variable to control whether to update the
`__version__` string with the Travis build number. If the build is
not a tag, it should be deployed to testpypi, with a build version.
If the build is a tag, it should be deployed to production pypi and
the version should be updated outside of Travis.

---
This change is [Reviewable](https://reviewable.io/reviews/plus3it/watchmaker/52)


KevinPlus3:
Reviewed 3 of 4 files at r1, 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/plus3it/watchmaker/52)*

ci/travis_pypi_setup.py
```diff
@@ -23,10 +22,12 @@
 
 GITHUB_REPO = 'plus3it/watchmaker'
 TRAVIS_CONFIG_FILE = os.path.join(
```
MarionTheBull: Why not remove all of this? Where is it being called? I think you removed all uses of it, remove 24 - 30 as well.

ci/travis_pypi_setup.py
```diff
@@ -9,7 +9,6 @@
 import json
```
MarionTheBull: is this being used?

ci/travis_pypi_setup.py
```diff
@@ -9,7 +9,6 @@
...
 import os
```
MarionTheBull: if you remove the unused TRAVIS_CONFIG_FILE you can also remove this.

ci/travis_pypi_setup.py
```diff
@@ -9,7 +9,6 @@
...
 from getpass import getpass
-import yaml
 from cryptography.hazmat.primitives.serialization import load_pem_public_key
```
MarionTheBull: unused

ci/travis_pypi_setup.py
```diff
@@ -9,7 +9,6 @@
...
 from cryptography.hazmat.backends import default_backend
```
MarionTheBull: unused

ci/travis_set_build.py
```diff
@@ -0,0 +1,84 @@
...
+    # __version__ = 'ver'
+    pattern = r"^(__version__ = ['\"])([^'\"]*)(['\"])"
+    repl = r"\g<1>\g<2>.dev{0}\g<3>".format(build)
+    version_file = os.path.join(PROJECT_ROOT, *file_paths)
+    print(
+        'Updating version in version_file "{0}" with build "{1}"'
+        .format(version_file, build)
+    )
+    replace(version_file, pattern, repl, flags=re.M)
+
```
MarionTheBull: Instead of using a one sided conditional to sys.exit the middle of a small main, why not break out the things into a function and use else to call it. This will also remove the dep on sys.

```

def append_file(build, file_paths):
    # The version line must have the form
    # __version__ = 'ver'
    pattern = r"^(__version__ = ['\"])([^'\"]*)(['\"])"
    repl = r"\g<1>\g<2>.dev{0}\g<3>".format(build)
    version_file = os.path.join(PROJECT_ROOT, *file_paths)
    print(
        'Updating version in version_file "{0}" with build "{1}"'
        .format(version_file, build)
    )
    replace(version_file, pattern, repl, flags=re.M)

def main(args):
    skip = args.skip
    build = args.build
    file_paths = args.file_paths

    if skip:
        print(
            'Not updating version for this build, `skip` set to "{0}"'
            .format(skip)
        )
    else:
        append_file(build, file_paths)
```

MarionTheBull: This could also be a little heavy handed when most of this could just be rolled into replace and the else statement could be just the replace.

MarionTheBull:
Reviewed 3 of 4 files at r1, 1 of 1 files at r2.
Review status: all files reviewed at latest revision, 5 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/plus3it/watchmaker/52)*

ci/travis_pypi_setup.py
lorengordon: Oops, good catch.

ci/travis_pypi_setup.py
lorengordon: I just checked and all the cryptography imports are being used...

ci/travis_set_build.py
lorengordon: I think the first suggestion is a good idea. `replace` is a nice, general purpose function that replaces any text in a file with any other text, so not a fan of the second.

Should I make the `pattern` and `repl` global defaults, with argparse params to override?

ci/travis_pypi_setup.py
MarionTheBull: why line is load_pem_public_key and default_backend on?  i cant find them outside of the removed lines?

ci/travis_set_build.py
MarionTheBull: i go both ways on this. making global makes it more obvious to update; however, it is specific to this function and i would assume rarely going to be changed for the project.  once we pick a direction any deviation to the naming convention would be a major change as it wouldn't be backwards compatible. thoughts?

ci/travis_pypi_setup.py
MarionTheBull: yea, i realized it was nesting expansions and i didnt expand them all.  my bad.

ci/travis_set_build.py
lorengordon: i think it's probably fine to leave them hard coded in the function in this case. maybe i'll just parameterize the `build_type`, which is hardcoded to 'dev' right now... also, pep 440 is a thing, enforced by pypi. who knew?

lorengordon: Odd that GitHub is not recognizing that this comment is now on an outdated diff. Implemented the former suggestion as discussed.

lorengordon:
Comments incorporated. Ready to review again.

KevinPlus3:
Reviewed 1 of 1 files at r3, 1 of 1 files at r4.
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/plus3it/watchmaker/52)*

MarionTheBull:
Reviewed 1 of 1 files at r3, 1 of 1 files at r4.
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/plus3it/watchmaker/52)*

PR #55, opened 2016-08-29
Remove_Redundancy

KevinPlus3:
Removing redundancy in argparse for logger setting.  `store_true` also sets the default value to `False`, so there is no need to include `Deafult=False`.

---
This change is [Reviewable](https://reviewable.io/reviews/plus3it/watchmaker/55)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -75,6 +75,10 @@ pyvenv*/
 # Installer logs
 pip-log.txt
 
+# Log files created by running program
+src/*.log
+tests/*.log
+
 # Unit test / coverage reports
 .coverage
 .tox
--- a/scripts/Watchmaker.py
+++ b/scripts/Watchmaker.py
@@ -11,7 +11,7 @@ if __name__ == "__main__":
                         help='Use S3 buckets instead of internet locations for files.')
     parser.add_argument('--config', dest='config', default='config.yaml',
                         help='Path or URL to the config.yaml file.')
-    parser.add_argument('--logger', dest='logger', action='store_true', default=False,
+    parser.add_argument('--logger', dest='logger', action='store_true',
                         help='Use stream logger for debugging.')
     parser.add_argument('--log-path', dest='log_path', default=None,
                         help='Path to the logfile for stream logging.')
--- a/src/watchmaker/__init__.py
+++ b/src/watchmaker/__init__.py
@@ -68,6 +68,7 @@ class Prepare(object):
         self.system_params = None
         self.system_drive = None
         self.execution_scripts = None
+        logging.basicConfig()
         self.logger = logging.getLogger()
 
         if arguments.logger and os.path.exists(arguments.log_path):


MarionTheBull:
Review status: 0 of 3 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[src/watchmaker/__init__.py, line 71 \[r2\]](https://reviewable.io:443/reviews/plus3it/watchmaker/55#-KQMKxRGwxl921WIjB0W:-KQMKxRHGi2cIo8oqQRY:b-u90b9i) ([raw file](https://github.com/plus3it/watchmaker/blob/2543e770fa999c513560a7fb8b8d5c5d12f66925/src/watchmaker/__init__.py#L71)):*
> ```Python
>         self.system_drive = None
>         self.execution_scripts = None
>         logging.basicConfig()
> ```

This should actually be used, we aren't actually defining things here as we should be.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/plus3it/watchmaker/55)*

MarionTheBull:
Reviewed 2 of 2 files at r1, 1 of 1 files at r2.
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/plus3it/watchmaker/55)*

KevinPlus3:
*[src/watchmaker/__init__.py, line 71 \[r2\]](https://reviewable.io:443/reviews/plus3it/watchmaker/55#-KQMKxRGwxl921WIjB0W:-KQMLhXg_TEQCGUDDfvD:b-hha9hs) ([raw file](https://github.com/plus3it/watchmaker/blob/2543e770fa999c513560a7fb8b8d5c5d12f66925/src/watchmaker/__init__.py#L71)):*
<details><summary><i>Previously, MarionTheBull (Marion Cobretti) wrote…</i></summary>
> This should actually be used, we aren't actually defining things here as we should be.
</details>
Right. This should be called before getLogger() unless your purpose is to not have any handlers attached to the logger.  I know we probably should construct the logging portion in abetter way - but figured that could be a different PR.  Right now, addressing this to help with testing.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/plus3it/watchmaker/55)*

MarionTheBull:
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[src/watchmaker/__init__.py, line 71 \[r2\]](https://reviewable.io:443/reviews/plus3it/watchmaker/55#-KQMKxRGwxl921WIjB0W:-KQMMhiAd_hKWA51kHMR:b-m1eryy) ([raw file](https://github.com/plus3it/watchmaker/blob/2543e770fa999c513560a7fb8b8d5c5d12f66925/src/watchmaker/__init__.py#L71)):*
<details><summary><i>Previously, KevinPlus3 (Kevin Hulsing) wrote…</i></summary>
> Right. This should be called before getLogger() unless your purpose is to not have any handlers attached to the logger.  I know we probably should construct the logging portion in abetter way - but figured that could be a different PR.  Right now, addressing this to help with testing.
</details>
Bah, you are right.  my bad.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/plus3it/watchmaker/55)*

PR #66, opened 2016-08-30
Writes join-domain and name-computer grains with salt linux worker

lorengordon:
*   Adds the join-domain grain to the salt linux worker. This enables support for the `oupath`, `adminusers`, and `admingroups` settings.
*   Adds the name-computer grain to the salt linux worker. This enables support for the `computername` configuration setting.
*   Modifies the approach to writing the join-domain grain for both linux and windows salt workers so that the resulting values will be interpreted correctly.

---
This change is [Reviewable](https://reviewable.io/reviews/plus3it/watchmaker/66)

### DIFF
--- a/src/watchmaker/workers/salt.py
+++ b/src/watchmaker/workers/salt.py
@@ -83,6 +83,7 @@ class SaltLinux(LinuxManager):
                 'formulaterminationstrings']
 
         self.sourceiss3bucket = self.config['sourceiss3bucket']
+        self.computername = self.config['computername']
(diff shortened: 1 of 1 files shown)


KevinPlus3:
Reviewed 1 of 1 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/plus3it/watchmaker/66)*

Issue #63, opened 2016-08-30
Reconsider log level for existing messages


lorengordon:
Some info messages should be debug, etc...

KevinPlus3:
Addressed with PR #69 
However, continue to keep your eyes open for future log messages and or if any have been missed.

Issue #60, opened 2016-08-30
`--logger --log-path=.` are not actually producing a log file


lorengordon:

KevinPlus3:
Addressed with PR #69

PR #69, opened 2016-08-31
Reworking Logger

KevinPlus3:
Added new private method for setting up logger. Refactored code and made some minor improvements. Also addresses issues #60 to #63.

---
This change is [Reviewable](https://reviewable.io/reviews/plus3it/watchmaker/69)

### DIFF
--- a/src/watchmaker/__init__.py
+++ b/src/watchmaker/__init__.py
@@ -67,13 +67,23 @@ class Prepare(object):
         logging.info('System Type: {0}'.format(self.system))
 
     def _prepare_logger(self, log_dir, log_file):
+        """
+        Prepares the logger for handling messages to a file and/or to stdout.
+        Args:
+            log_dir (str):
+                Path of a directory. If this object is not None, then this
+                directory will be used to store a log file.
+            log_file (str):
+                Path to a file. If this object is not None, then this path
+                to a file will be used as the log file.
+        """
         log_filename = None
         if log_dir and os.path.exists(log_dir):
             if os.path.isfile(log_dir):
                 log_msg = '{0} is a file and not a directory.'.format(log_dir)
                 log_type = 'error'
             else:
-                log_filename=os.path.join(
+                log_filename = os.path.join(
                     log_dir,
                     'watchmaker-{0}.log'.format(str(datetime.date.today()))
                 )
--- a/src/watchmaker/__init__.py
+++ b/src/watchmaker/__init__.py
@@ -50,7 +50,7 @@ class Prepare(object):
             logger (bool):
                 Enables self.logger to a file.
             log_path (str):
-                Path to logfile for stream self.logger.
+                Path to log directory for stream self.logger.
         """
         self.kwargs = {}
         self.noreboot = arguments.noreboot
@@ -67,6 +67,7 @@ class Prepare(object):
         self._prepare_logger(arguments.log_path)
         self.logger.info('Parameters:  {0}'.format(self.kwargs))
         self.logger.info('System Type: {0}'.format(self.system))
+        sys.exit()
 
     def _prepare_logger(self, log_path):
         if log_path and os.path.exists(log_path):
@@ -238,8 +239,6 @@ class Prepare(object):
 
         After execution the system should be properly provisioned.
         """
-        self.logger.info('+' * 80)
-
         self._get_system_params()
         self.logger.info(self.system_params)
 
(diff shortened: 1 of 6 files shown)


src/watchmaker/__init__.py
```diff
@@ -108,30 +135,35 @@ def _get_config_data(self):
...
-                      'work!\nPlease make sure your config is available.')
+                logging.critical(
+                    'The URL used to get the user config.yaml file did not '
+                    'work!\nPlease make sure your config is available.'
+                )
                 sys.exit(1)
 
         if self.config_path and not os.path.exists(self.config_path):
-            self.logger.warning(
+            logging.warning(
```
lorengordon: I feel like this should probably be a critical error, log and exit. As is, getting here means the user passed in a param, the param is not validating properly, but we are proceeding anyway with a config different than they intended. That seems counter to the principle of least surprise.

src/watchmaker/__init__.py
```diff
@@ -108,30 +135,35 @@ def _get_config_data(self):
...
             data = f.read()
 
         if data:
             self.config = yaml.load(data)
         else:
-            logging.debug('No data to load.')
+            logging.critical(
+                'Unable to load the data of the default or'
+                ' the user supplied config.'
+            )
```
lorengordon: I think the only way to get into this branch is if the file is empty. This should probably exit if there is no config data, right? Does `logging.critical()` automatically exit?

KevinPlus3: Yeah. I was asking Adam about that earlier when I was like shouldn't this be critical.  `logging.critical` does not automatically exit (I don't think).  But I think we may need to do some refactoring in terms of properly exiting.  I can add a sys.exit(1) as well for now.

src/watchmaker/__init__.py
```diff
@@ -51,42 +47,73 @@ def __init__(self, arguments):
...
+        elif log_file:
+            if os.path.isdir(log_file):
+                log_msg = '{0} is a directory and not a file.'.format(log_file)
+                log_type = 'error'
+            else:
+                log_filename = log_file
+                log_msg = 'Start time: {0}'.format(datetime.datetime.now())
+                log_type = 'info'
+        elif not log_dir and not log_file:
+            log_msg = 'Stream logger is not enabled!'
```
lorengordon: I don't think users are going to know what a stream logger is. (I still don't....) What does this really mean, from the user's perspective?

src/watchmaker/__init__.py
```diff
@@ -108,30 +135,35 @@ def _get_config_data(self):
                     shutil.copyfileobj(response, outfile)
                 self.config_path = 'config.yaml'
             except urllib.error.URLError:
-                print('The URL used to get the user config.yaml file did not '
-                      'work!\nPlease make sure your config is available.')
+                logging.critical(
+                    'The URL used to get the user config.yaml file did not '
+                    'work!\nPlease make sure your config is available.'
```
lorengordon: Thoughts on moving away from multiline messages, now that we've moved to the logging facility? I think they make the log file ugly. If we need multiple lines, we maybe should use multiple `logging()` calls.

src/watchmaker/cli.py
```diff
@@ -15,15 +15,16 @@ def main():
...
-    parser.add_argument('--logger', dest='logger', action='store_true',
-                        help='Use stream logger for debugging.')
-    parser.add_argument('--log-path', dest='log_path', default=None,
-                        help='Path to the logfile for stream logging.')
     parser.add_argument('--saltstates', dest='saltstates', default=None,
                         help=(
                             'Define the saltstates to use.  Must be None, '
                             'Highstate, or comma-seperated-string'
                         ))
+    group = parser.add_mutually_exclusive_group(required=False)
```
lorengordon: Nice, was wondering how we were going to do that, turns out it's just part of argparse. Looking down the road, maybe name this `log_group`, so if we have another set of mutually exclusive params the naming continues to make sense?

lorengordon:
A few comments, but looks good overall.

src/watchmaker/__init__.py
KevinPlus3: Probably nothing to a user at the top level.  I'll shorten to just logger.

src/watchmaker/__init__.py
KevinPlus3: I agree with you on this front.  If we want it to be similar to typical logs, then most if not all actually keep info on one line per log.

src/watchmaker/cli.py
KevinPlus3: :thumbs_up: Will do.

lorengordon:
Reviewed 3 of 5 files at r1, 2 of 2 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/plus3it/watchmaker/69)*

PR #71, opened 2016-09-01
LogHandler Class

KevinPlus3:
Moved a lot of the logging functionality to the LogHandler class in the exceptions section.  All of the methods are static since logging is a global object.  Set up in preparation for handling -vvv arguments where we can have both a stdout and a file for logging to, with different logging levels if needed (and different formats too).  Moved sys.exit() to the add method so this can be the place where we will handle logs and handle critical and error log types.

---
This change is [Reviewable](https://reviewable.io/reviews/plus3it/watchmaker/71)


src/watchmaker/exceptions/__init__.py
```diff
@@ -1,18 +1,127 @@
+import datetime
 import logging
+import os
 import sys
 
 
-def SystemFatal(msg):
-    """
-    This is a temporary exception handler for the project.
+class LogHandler:
```
lorengordon: It feels a little odd for this to be in exceptions. Maybe it's own module, so the import semantics are just `from watchmaker import LogHandler as log`? Also, is there some way to make the `add()` function the "default" for the class, so the usage semantics are just `log(msg, log_type, exc)`?

KevinPlus3: I agree .... I'll move it over to its own module and look into the other good suggestions as well.

lorengordon: My thinking on why it is odd to me is that with this in the exceptions module, I would expect to use `raise LogHandler()`, but that doesn't seem like what we want to do.

src/watchmaker/workers/yum.py
```diff
@@ -1,8 +1,7 @@
 import json
-import logging
 import re
-import sys
 
+from watchmaker.logger import LogHandler
```
lorengordon: Can we use the `log()` semantics in this module as well?

KevinPlus3: Frak!  LOL Yes

lorengordon:
Reviewed 1 of 6 files at r1, 5 of 6 files at r2, 1 of 1 files at r3.
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/plus3it/watchmaker/71)*

PR #78, opened 2016-10-18
Refactored salt.py

KevinPlus3:
Added a new SaltBase class, SaltLinux will "inherit" from SaltBase and LinuxManager, while SaltWindows will "inherit" from SaltBase and WindowsManager.  Tested on RHEL6 instance from Tom's AMIgen run and tested on Windows 2012 R2 Server.  Reduces duplication of codes between SaltLinux and SaltWindows.


KevinPlus3:
Want to test this out more.

PR #80, opened 2016-10-20
Refactor Salt.py

KevinPlus3:
A lot of reduction in code. Now has a SaltBase from which SaltLinux and SaltWindows inherit. Modified the way we import boto3 and reduced call for importing to one.  Removed abstraction of ManagerBase to greatly reduce call of super in LinuxManager and WindowsManager.  Extension of ManagerBase seemed to be a better fit.  Fixed issue with sourceiss3bucket cli parameter that was not overriding config.yaml value.  Tested on Tom's CentOS6 AMI.


KevinPlus3:
Closing and re-opening to re-run Travis.

src/watchmaker/workers/salt.py
```diff
@@ -7,195 +7,157 @@
...
+            self.salt_base_env,
+            self.salt_formula_root
+        ]:
             try:
-                os.makedirs(saltdir)
+                os.makedirs(salt_dir)
             except OSError:
-                if not os.path.isdir(saltdir):
+                if not os.path.isdir(salt_dir):
                     raise
```
MarionTheBull: naked raise should be cleaned up. this i think was my bad, as we hadnt had a clear path on exception handling, unless its always being caught.

PR #103, opened 2017-01-15
Moves sys.exit() to the outermost edge of the cli

lorengordon:
Previously, wm_exit() was calling sys.exit() from the exceptions module, which causes watchmaker to behave a bit poorly when watchmaker itself is imported as a library. Watchmaker would force the program to exit, rather than letting the program catch errors and handle its own exit.

Moving sys.exit() to the cli module places the exit at the outermost edge of the watchmaker code, and uses python's raise and exception handling to determine the cli exit code. i.e. if a traceback occurs, the cli exits with 1, otherwise it exits 0.

### DIFF
--- a/src/watchmaker/__init__.py
+++ b/src/watchmaker/__init__.py
@@ -10,7 +10,7 @@ import yaml
 from six.moves import urllib
 
 from watchmaker import static
-from watchmaker.exceptions import ExcLevel, wm_exit
+from watchmaker.exceptions import WatchmakerException
 from watchmaker.managers.workers import (LinuxWorkersManager,
                                          WindowsWorkersManager)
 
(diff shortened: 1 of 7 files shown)


src/watchmaker/__init__.py
```diff
@@ -251,15 +257,16 @@ def install_system(self):
                 self.saltstates
             )
         else:
-            wm_exit('There is no known System!', ExcLevel.Critical, False)
+            msg = 'There is no known System!'
+            plog.critical(msg)
+            raise WatchmakerException
```
KevinPlus3: Is this supposed to be WatchmakerException(msg) like the others?

lorengordon: yes 😭

lorengordon: all better!

PR #106, opened 2017-01-16
Moves loggers back into class functions

lorengordon:
We had been doing this previously, but using `self.log` felt a little weird, so we defined the loggers globally instead. That is turning out to feel even more weird, with multiple loggers per module (when there are multiple classes logging in one module), and the need to pass the right logger around between different functions.

This changes things back so that watchmaker uses `self.log` when logging within a class. The logger only needed to be defined in two places, Prepare() and ManagerBase(), since all managers and workers inherit from ManagerBase.

Within a class, the logger name is set to include both the module path and the class name:

```
'{0}.{1}'.format(__name__, self.__class__.__name__)
```

The log format is updated as well, so that the logger name and the process number are in the output:

```
%(asctime)s [%(name)s][%(levelname)-5s][%(process)s]: %(message)s
```

The combination results in messages formed like this:

```
2017-01-16 19:15:32,996 [watchmaker.Prepare][INFO ][10084]: Got scripts to execute: ['Yum', 'Salt'].
2017-01-16 19:15:32,998 [watchmaker.managers.base.Yum][DEBUG][10084]: Destination: /etc/yum.repos.d/systemprep-repo-salt-el6.repo
2017-01-16 19:15:36,403 [watchmaker.managers.base.SaltLinux][DEBUG][10084]: /usr/tmp/saltinstall0UD1BH being cleaned up.
2017-01-16 19:15:36,404 [watchmaker.managers.base.SaltLinux][INFO ][10084]: Removed temporary data in working directory -- /usr/tmp/saltinstall0UD1BH
2017-01-16 19:15:36,404 [watchmaker.managers.base.SaltLinux][INFO ][10084]: Exiting cleanup routine...
2017-01-16 19:15:36,404 [watchmaker.Prepare][INFO ][10084]: Stop time: 2017-01-16 19:15:36.404391
```

In addition, there was a bug in the verbosity parameter, such that the count action was not working correctly. That is also fixed.

### DIFF
--- a/src/watchmaker/__init__.py
+++ b/src/watchmaker/__init__.py
@@ -15,7 +15,6 @@ from watchmaker.managers.workers import (LinuxWorkersManager,
                                          WindowsWorkersManager)
 
 __version__ = '0.0.1'
-plog = logging.getLogger('Prepare')
 
 
 class PrepArguments(object):
@@ -52,6 +51,9 @@ class Prepare(object):
             log_dir (str) or log_file (str):
                 Path to log directory or file for logging.
         """
+        self.log = logging.getLogger(
+            '{0}.{1}'.format(__name__, self.__class__.__name__)
+        )
         self.kwargs = {}
         self.noreboot = arguments.noreboot
         self.s3 = arguments.sourceiss3bucket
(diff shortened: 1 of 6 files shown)


src/watchmaker/__init__.py
```diff
@@ -52,6 +51,9 @@ def __init__(self, arguments):
             log_dir (str) or log_file (str):
                 Path to log directory or file for logging.
         """
+        self.log = logging.getLogger(
+            '{0}.{1}'.format(__name__, self.__class__.__name__)
+        )
```
KevinPlus3: Is this one necessary if you have it in the __init__ function of WatchmakerException?

PR #144, opened 2017-02-20
Uses flake8 plugins and pylint for static code analysis

lorengordon:


KevinPlus3:
I assume it follows some convention to use s3_ and fh_, that is to say, to end the variable name with an underscore?

lorengordon:
Yeah, convention to pad a variable name to three characters. The pylint [bad-name](http://pylint-messages.wikidot.com/messages:c0103) regex for variables requires three characters. So, alternatively, could also think of a three-character name.

PR #145, opened 2017-02-22
Removes Parameters key from config file

lorengordon:


src/watchmaker/__init__.py
```diff
@@ -286,11 +284,11 @@ def _get_config(self):
...
                 if not config[worker_name].get('__merged'):
                     # Merge worker_args into config params
-                    config[worker_name]['Parameters'].update(self.worker_args)
+                    config[worker_name]['config'].update(self.worker_args)
                     config[worker_name]['__merged'] = True
-            except KeyError:
+            except Exception:
                 msg = (
-                    'Malformed worker, missing "Parameters" key; worker = {0}'
+                    'Failed to merge worker config; worker={0}'
```
KevinPlus3: From last PR, I noticed that you changed any {}.format to using %s - so I assume you want consistency and just missed this one.

lorengordon: That change was only when the message string was being formatted in the `log()` statement. This is formatted separately, so `msg` can be reused.

PR #151, opened 2017-02-28
Updates powershell bootstrapper and adds userdata examples to usage doc

lorengordon:


docs/installation.md
```diff
@@ -18,16 +18,10 @@ can guide you through the process.
...
 
 ```bash
-git clone https://github.com/plus3it/watchmaker.git
-```
-
-Or download the [`tarball`][3]:
-
-```bash
-curl -OL https://github.com/plus3it/watchmaker/tarball/develop
+git clone https://github.com/plus3it/watchmaker.git && cd watchmaker
```
KevinPlus3: You don't mention the "git submodule update --init --recursive" command here. Should that be added?

lorengordon: That was already present. Did I delete it by accident?

KevinPlus3: D'oh. I should have expanded the collapsed lines and I would've noticed it.  My bad!

PR #224, opened 2017-03-18
Avoids duplicating stdout/stderr in memory

lorengordon:
Previous approach using splitlines doubled the memory requirement,
which was very inefficient particularly for highstate execution as
there is a lot of output.

This patch uses io.BytesIO to treat the string like a file object,
which avoids reading it into memory all at once.


lorengordon:
Closing in favor of #225

PR #225, opened 2017-03-18
Streams stdout/stderr to logger using threads

lorengordon:
subprocess.communicate() blocks and reads all the output into
memory before returning. It is safe in that it uses threads to
avoid deadlocks when capturing both stdout and stderr in a pipe.

This patch uses threads to read the stdout and stderr pipes, and
to log the output as it is generated. This approach continues to
avoid deadlocks, but also avoids blocking and reading all the output
into memory.

### DIFF
--- a/src/watchmaker/managers/base.py
+++ b/src/watchmaker/managers/base.py
@@ -186,8 +186,11 @@ class ManagerBase(object):
 
     @staticmethod
     def _pipe_logger(pipe, logger, prefix_msg=''):
-        for line in iter(pipe.readline, b''):
-            logger('%s%s', prefix_msg, line.rstrip())
+        try:
+            for line in iter(pipe.readline, b''):
+                logger('%s%s', prefix_msg, line.rstrip())
+        finally:
+            pipe.close()
 
     def call_process(self, cmd):
         """
@@ -223,8 +226,6 @@ class ManagerBase(object):
 
         stdout_reader.join()
         stderr_reader.join()
-        process.stdout.close()
-        process.stderr.close()
 
         returncode = process.wait()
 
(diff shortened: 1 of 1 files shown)


lorengordon:
Yes, it's been tested on centos6, centos7, ws2012r2, and ws2016. I try to run all my branches on those systems, using the [userdata at the end of the Contributing doc](https://github.com/plus3it/watchmaker/blob/develop/CONTRIBUTING.md) to make it easier.

The code here is derived from the code used by `subprocess._communicate()`, which also does not use queues. The main difference is that `_communicate()` blocks to read all the data, and we loop over the pipe to log it immediately as it comes in.  The "call to process" is `_pipe_logger()`, specified as the target of `thread.Thread()`... There is no sharing of state or data between the two threads, so a queue seems like overkill, but I'm pretty new to this so could easily be missing something.

lorengordon:
Also, the python logger is [intended to be thread-safe](https://docs.python.org/2/library/logging.html#thread-safety)...

src/watchmaker/managers/base.py
```diff
@@ -202,14 +208,27 @@ def call_process(self, cmd):
...
+
+        stderr_reader = threading.Thread(
+            target=self._pipe_logger,
+            args=(process.stderr, self.log.error, 'Command stderr: '))
+        stderr_reader.daemon = True
+        stderr_reader.start()
+
+        stdout_reader.join()
+        stderr_reader.join()
+        process.stdout.close()
```
lorengordon: I could move the closing of the pipe into `_pipe_logger()`, after the loop, which would better match how the pipe/thread is managed in `subprocess._communicate()`. That way the handle is closed already when the thread returns.

PR #236, opened 2017-04-04
Supports user_formulas as a map of `name:url`

### LINKED ISSUE #163: Rework `user_formulas` to support a map of `{formula_name}: {formula_source}`
Right now, effectively, we use `formula_termination_strings` to determine the formula name (and thus the name of the formula directory in the <salt_srv> root). That gets to be rather hacky. Here's the current config definition:

```
      formula_termination_strings:
        - -master
        - -latest
      user_formulas:
        #To add other formulas, make sure it is a url to a zipped file as follows:
        #- https://s3.amazonaws.com/salt-formulas/systemprep-formula-master.zip
        #- https://s3.amazonaws.com/salt-formulas/ash-linux-formula-master.zip
        #To "overwrite" submodule formulas, make sure name matches submodule names.
```

 Instead, change the spec for `user_formulas` to be a map of the formula name and the formula source:

```
      user_formulas:
        #systemprep-formula: https://s3.amazonaws.com/salt-formulas/systemprep-formula-master.zip
        #ash-linux-formula: https://s3.amazonaws.com/salt-formulas/ash-linux-formula-master.zip
```

This would deprecate `formula_termination_strings`.

lorengordon:
Fixes #163

### DIFF
--- a/src/watchmaker/workers/salt.py
+++ b/src/watchmaker/workers/salt.py
@@ -43,18 +43,13 @@ class SaltBase(ManagerBase):
             - ``none``: Do not apply any salt states.
             - ``highstate``: Apply the salt "highstate".
 
-        user_formulas (:obj:`list`):
-            (Defaults to ``[]``) URLs to archives (zip files) of salt formulas
-            that will be downloaded and extracted on the system.
-        formula_termination_strings (:obj:`list`):
-            (Defaults to ``[]``) Strings that should be removed from the end
-            of the salt formulas. For example, when downloading the zip archive
-            from GitHub, the name of the branch is appended to the archive.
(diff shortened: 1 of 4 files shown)


src/watchmaker/workers/salt.py
```diff
@@ -44,17 +44,12 @@ class SaltBase(ManagerBase):
             - ``highstate``: Apply the salt "highstate".
 
         user_formulas (:obj:`list`):
```
KevinPlus3: Should this now say "user_formulas (:obj:'dict'):"?

lorengordon: Yep, fixed.

PR #284, opened 2017-05-02
Uses click for cli argument processing

lorengordon:
Replaces argparse. The argparse library dependency was required to
use argparse on py26, but the libary is not actively maintained.
Argparse is actively maintained as a core python library, but only
available in py27 and later. Installing argparse from pypi can
result in some odd behavior on imports, as we may get the core
library _or_ the pypi library (depending on platform, typically).

Using click ensures consistent behavior across all platforms and
python versions.

I looked at `invoke` also, but couldn't figure out how to replicate some of our option types, and felt the docs weren't quite as mature as `click`. This implementation with `click` results in identical cli usage as the prior approach with `argparse`.

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -9,57 +9,51 @@ the output of `watchmaker --help`, showing the CLI options.
 
 ```shell
 # watchmaker --help
-usage: watchmaker [-h] [-v] [-c CONFIG_PATH] [-l LOG_LEVEL] [-d LOG_DIR] [-n]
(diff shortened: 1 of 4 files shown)


lorengordon:
Opening for discussion. Still need to test on Windows.

KevinPlus3:
I am reading through on click currently.  I certainly have no objections to switching over to a tool that is more in line with what we need for Watchmaker especially since argparse tends to cause odd behavior on installations from pypi.

PR #320, opened 2017-06-04
Includes failed state output in exception message

lorengordon:
Previously, the exception message was very generic as it was raise by the manager method, `call_process`, which isn't aware of the context of the call.

This patch moves the raise to the salt worker method, `process_states`, which allows us to be more specific about the error and include more information about the failure. Now, when a state fails, the last messages in the log file will be referencing exactly the states that failed. This _ought_ to result in more accurate issue/bug reports.

### DIFF
--- a/src/watchmaker/workers/salt.py
+++ b/src/watchmaker/workers/salt.py
@@ -440,7 +440,7 @@ class SaltBase(ManagerBase):
                 )
                 cmd.extend(['state.sls', states])
 
-            ret = self.run_salt(cmd, raise_error=False)
+            ret = self.run_salt(cmd, log_pipe='stderr', raise_error=False)
 
             if ret['retcode'] != 0:
                 failed_states = self._get_failed_states(
--- a/src/watchmaker/workers/salt.py
+++ b/src/watchmaker/workers/salt.py
@@ -420,25 +420,23 @@ class SaltBase(ManagerBase):
             self.log.info(
                 'No States were specified. Will not apply any salt states.'
             )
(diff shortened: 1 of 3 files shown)


lorengordon:
The output on a full run isn't quite what I was hoping for. Still playing with the options on this one...

lorengordon:
@KevinPlus3 ok, this is good for now. runs correctly on windows and linux, py26, py27, and py36. please take another look tomorrow or whenever and let me know if it makes sense or not.

src/watchmaker/workers/salt.py
```diff
@@ -315,8 +334,8 @@ def service_stop(self, service):
             'service.stop', service,
             '--out', 'newline_values_only'
         ]
-        ret = self.run_salt(cmd, stdout=True)
-        return ret.strip().lower() == b'true'
+        ret = self.run_salt(cmd)
+        return ret['stdout'].strip().lower() == b'true'
```
KevinPlus3: Combine these two into one line? Similar to lines 315/316?

src/watchmaker/workers/salt.py
```diff
@@ -336,8 +355,8 @@ def service_start(self, service):
             'service.start', service,
             '--out', 'newline_values_only'
         ]
-        ret = self.run_salt(cmd, stdout=True)
-        return ret.strip().lower() == b'true'
+        ret = self.run_salt(cmd)
+        return ret['stdout'].strip().lower() == b'true'
```
KevinPlus3: Also combine these two into one line as well?

Issue #322, opened 2017-06-07
using py3 on linux not working


ewierschke:
Using either CentOS 6.9 or 7.3 with python34 installed from repo produces the following on watchmaker execution.

~~~
Traceback (most recent call last):
  File "/usr/bin/watchmaker", line 11, in <module>
    sys.exit(main())
  File "/usr/lib/python3.4/site-packages/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "/usr/lib/python3.4/site-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/usr/lib/python3.4/site-packages/click/core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/lib/python3.4/site-packages/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "/usr/lib/python3.4/site-packages/watchmaker/cli.py", line 93, in main
    sys.exit(watchmaker_client.install())
  File "/usr/lib/python3.4/site-packages/watchmaker/__init__.py", line 415, in install
    workers_manager.worker_cadence()
  File "/usr/lib/python3.4/site-packages/watchmaker/managers/workers.py", line 27, in worker_cadence
    **configuration
  File "/usr/lib/python3.4/site-packages/watchmaker/workers/yum.py", line 43, in __init__
    self.dist_info = self.get_dist_info()
  File "/usr/lib/python3.4/site-packages/watchmaker/workers/yum.py", line 59, in get_dist_info
    with open(name='/etc/system-release', mode='rb') as fh_:
TypeError: Required argument 'file' (pos 1) not found
~~~

KevinPlus3:
This does not sound like a Python issue - but rather an issue with opening `/etc/system-release`.  Are you saying this working fine in Python 2.*, but did not work fine with Python 3.*?

KevinPlus3:
Adding comment from Loren as he has researched it - "It is a py3 issue.  py3 switched the default `open` to `io.open`, which changed the name of the first argument from `name` in py2 to `file`."

KevinPlus3:
Once this is addressed, then a new error pops up:
```
Traceback (most recent call last):
  File "/usr/bin/watchmaker", line 11, in <module>
    load_entry_point('watchmaker==0.4.5.dev0', 'console_scripts', 'watchmaker')()
  File "/usr/lib/python3.4/site-packages/click/core.py", line 722, in __call__
[...]

PR #323, opened 2017-06-07
Fixing issues with `open`.

KevinPlus3:
py2's `open` uses `name` as key for parameter while `py3` uses `file`.  Both expect path to be in first position, so a potential easy fix to just remove the key `name=`.

Fixes issue #322


lorengordon:
:shipit:

PR #330, opened 2017-06-26
Output Watchmaker Version

KevinPlus3:
Adding debug line to output Watchmaker version being run.

Addresses issue #329

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -80,6 +80,7 @@ src/*.log
 tests/*.log
 
 # Unit test / coverage reports
+.cache
 .coverage
 .tox
 .coverage.*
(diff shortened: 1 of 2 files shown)


lorengordon:
Probably worth adding `.cache` to `.gitignore`...

KevinPlus3:
Removed .cache directory.  Not sure if I put the lines in the right section as I was playing around with testing at that point.

.gitignore
```diff
@@ -80,9 +80,12 @@ src/*.log
 tests/*.log
 
 # Unit test / coverage reports
+.cache/
```
lorengordon: just `.cache` will do it

.gitignore
```diff
@@ -80,9 +80,12 @@ src/*.log
...
 .coverage
 .tox
 .coverage.*
+__pycache__/
+*/__pycache__/
```
lorengordon: Is git attempting to include these `__pycache__` directories now? I haven't seen that yet...

Issue #321, opened 2017-06-04
Start Incorporating Tests


KevinPlus3:
We may need to start the process of expanding Watchmaker to include tests, similar in vein to what Salt does, so we can start testing the results and errors we may expect from Watchmaker.  This will help to improve our review/approval process.

KevinPlus3:
Added logger tests in #333.

lorengordon:
Might want to switch from pytest-capturelog to [pytest-catchlog](https://pypi.python.org/pypi/pytest-catchlog). pytest-capturelog hasn't been updated in a good long while and does not appear to really support py3. pytest-catchlog is still active, and (skimming the issues) seems to have support for merging into pytest-core...

KevinPlus3:
Cool. Will do once I get back to adding more tests ...

PR #333, opened 2017-06-27
Adding Logger Tests

KevinPlus3:
Obtain coverage for `__init__.py` in logger.  Start of adding tests to issue #321.

### DIFF
--- a/tests/test_logger.py
+++ b/tests/test_logger.py
@@ -1,0 +1,80 @@
+# -*- coding: utf-8 -*-
+"""Watchmaker main test module."""
+from __future__ import (absolute_import, division, print_function,
+                        unicode_literals, with_statement)
+
+import logging
+import os
+import shutil
+import sys
+
+import pytest
+
+from watchmaker.logger import (LOG_LEVELS, exception_hook, make_log_dir,
+                               prepare_logging)
+
+
+def test_log_level_dict():
+    """Tests the LOG_LEVELS dict."""
+    # Test strings that are in the dict.
+    level = LOG_LEVELS['critical']
+    assert level == logging.CRITICAL
+    level = LOG_LEVELS['debug']
+    assert level == logging.DEBUG
+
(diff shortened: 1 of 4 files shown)


tests/test_logger.py
```diff
@@ -0,0 +1,86 @@
...
+from watchmaker.logger import exception_hook
+from watchmaker.logger import LOG_LEVELS
+from watchmaker.logger import make_log_dir
+from watchmaker.logger import prepare_logging
+
+import logging
+import os
+import shutil
+import sys
+import unittest
```
lorengordon: isort order is builtin imports, then third-party imports, then application imports. it's also ordered alphabetically (and case matters). switch this around to:

```
import logging
import os
import shutil
import sys
import unittest

from watchmaker.logger import LOG_LEVELS
from watchmaker.logger import exception_hook
from watchmaker.logger import make_log_dir
from watchmaker.logger import prepare_logging
```

tests/test_logger.py
```diff
@@ -0,0 +1,88 @@
...
+import shutil
+import sys
+
+from watchmaker.logger import (LOG_LEVELS, exception_hook, make_log_dir,
+                               prepare_logging)
+
+if sys.version_info < (2, 7):
+    import unittest2 as unittest
+else:
+    import unittest
```
lorengordon: How about using `six` for compatibility stuff?

https://pythonhosted.org/six/#unittest-assertions

KevinPlus3: Doesn't seem to support it.  `six` doesn't seem to have a module rename for unittest and the support for assertions is only supported for Python 2.7+, so we would be stuck in the same boat for Python 2.6.  :(

KevinPlus3: And `six` at this moment does not support the `assertLogs` assertion.  No doubt lots of people would love that feature.

lorengordon: Bah.

PR #342, opened 2017-07-06
Manages selinux state around salt execution

### LINKED ISSUE #341: Manage selinux state
As caught per Tom, "executing watchmaker via init script (userdata or otherwise) needs to be done with `setenforce 0` being executed first (just like firewall-cmd)", especially for EL7 and running Watchmaker via UserData.

lorengordon:
* Tests if selinux is enforcing prior to salt state execution
* If selinux is enforcing, set it to permissive mode
* After salt completes, if selinux was enforcing previously,
  set it back to enforcing (even if salt raised an exception)

Fixes #341


lorengordon:
@KevinPlus3 this is passing our internal tests, any concerns?

PR #358, opened 2017-08-02
Fix Pip install for FIPS compliance

WilliamsPaulM:
Altered CFN to match the options used in the documentation: http://watchmaker.readthedocs.io/en/stable/usage.html?highlight=fips#watchmaker-as-ec2-userdata

Launching EL7 SPEL will fail without FIPS compliance


lorengordon:
Good call. I missed that one when I created these. Can you make the same edit to the [autoscaling template](https://github.com/plus3it/watchmaker/blob/develop/docs/files/cfn/templates/watchmaker-lx-autoscale.template#L1003)?

WilliamsPaulM:
Yes sir

PR #359, opened 2017-08-02
Fix Pip install for FIPS compliance - ASG

WilliamsPaulM:
Altered CFN to match the options used in the documentation: http://watchmaker.readthedocs.io/en/stable/usage.html?highlight=fips#watchmaker-as-ec2-userdata

Launching EL7 SPEL will fail without FIPS compliance

RE: https://github.com/plus3it/watchmaker/pull/358#issuecomment-319724084


lorengordon:
Oh, ok, we can do it this way. But generally there is no need to create new PRs for related work. With git and PRs, you can just add or amend a commit on a branch with an open PR, and Github will include it in the existing PR.

WilliamsPaulM:
Ah, sorry, my experience with Github is limited....I'll remember that for next time.

PR #369, opened 2017-08-12
Applies updates and standards across cfn templates

lorengordon:
* Includes `appscript` workaround for FIPS-enabled AMIs.  Script must be S3-hosted and KMS-encrypted, and the instance must be assigned an instance role with appropriate permissions to retrieve and decrypt the object.
* Updates all templates so that `appscript` must be S3-hosted (if used). Improves security posture for applications.
* Corrects usage of several watchmaker options, as the options were renamed at some point.
* Eliminates a deprecation warning when calling pip install.

### DIFF
--- a/docs/files/cfn/templates/watchmaker-lx-autoscale.template
+++ b/docs/files/cfn/templates/watchmaker-lx-autoscale.template
@@ -218,6 +218,12 @@
             "Type" : "String",
             "Default" : ""
         },
+        "WatchmakerAdminUsers" :
+        {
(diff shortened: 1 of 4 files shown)


lorengordon:
@WilliamsPaulM This PR includes a workaround so that you can use the AppScriptUrl parameter with a FIPS-enabled AMI. There are some requirements: 
> Includes appscript workaround for FIPS-enabled AMIs. Script must be S3-hosted and KMS-encrypted, and the instance must be assigned an instance role with appropriate permissions to retrieve and decrypt the object.

PR #409, opened 2017-11-01
Add terraform templates

userhas404d:
- Provide terraform templates that enable the existing CloudFormation templates to be deployable via Terraform
- Moved CloudFormation templates their own respective directories to support terraform
- Updated `docs\usage.md` with terraform template overview and URLs


docs/usage.md
```diff
@@ -176,16 +176,56 @@ maps", and provide one for each of the templates above.
...
+    files.
+```
+
+Variable values can be input interactively via the terraform console or 
+via a Terraform module. An example Terraform module that calls the 
+lx-autoscale template is shown below.
+
+```
+module "test-lx-instance" {
+  source = "./lx-autoscale"
```
lorengordon: Let's set the source to `github.com/plus3it/watchmaker//docs/files/templates/lx-autoscale`. See https://www.terraform.io/docs/modules/sources.html#github. This way no one actually _has_ to download the files locally.

PR #411, opened 2017-11-05
Corrects alignment for rst bulleted syntax

lorengordon:


lorengordon:
[Current rendering](http://watchmaker.readthedocs.io/en/develop/usage.html#watchmaker-as-a-terraform-template):

image

Updated rendering:

image

PR #419, opened 2017-11-19
Resolution to Issue 418

KevinPlus3:
Adding argument for exlude_states to salt.py. Inserting 'exclude=' into processing of states.

### DIFF
--- a/src/watchmaker/workers/salt.py
+++ b/src/watchmaker/workers/salt.py
@@ -51,6 +51,10 @@ class SaltBase(ManagerBase):
             - ``none``: Do not apply any salt states.
             - ``highstate``: Apply the salt "highstate".
 
+        exclude_states: (:obj:`str`)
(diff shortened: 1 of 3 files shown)


src/watchmaker/workers/salt.py
```diff
@@ -99,6 +99,7 @@ def __init__(self, *args, **kwargs):
         self.admin_groups = kwargs.pop('admin_groups', None) or ''
         self.admin_users = kwargs.pop('admin_users', None) or ''
         self.salt_states = kwargs.pop('salt_states', None) or ''
+        self.exclude_states = kwargs.pop('exclude_states', None) or ''
```
lorengordon: Need to add the new arg to the docstring on the `SaltBase` class

lorengordon:
Also should add the arg to the config doc, https://github.com/plus3it/watchmaker/blob/develop/docs/configuration.md#salt

PR #450, opened 2017-12-07
Pins earlier html5lib version to retain compatibility with bleach

lorengordon:
See https://github.com/mozilla/bleach/issues/337


lorengordon:
The mozilla/bleach team appears to be moving quickly on this, so may not need to merge...

lorengordon:
Bleach was patched and a new version released already, no need to merge this.

PR #464, opened 2017-12-14
Added Info for Azure Usage

KevinPlus3:
Added brief information in Usage documentation on using Custom Script Extension to run bash/PowerShell scripts for Watchmaker on an Azure virtual machine.

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -57,7 +57,9 @@ Options:
   --help                          Show this message and exit.
 ```
 
-## `watchmaker` as EC2 userdata
+## `watchmaker` in AWS
+
+### `watchmaker` as EC2 userdata
 
 Calling watchmaker via EC2 userdata is a variation on using it as a CLI
 utility. The main difference is that you must account for installing watchmaker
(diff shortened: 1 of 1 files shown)


eemperor:
@KevinPlus3 The block of text was inserted in between the heading `watchmaker as EC2 userdata` and it's subheading `Linux`.  Should be good once you move it outside in its own section.

KevinPlus3:
Done.  Good catch.  I had difficulty deciding whether or should immediately follow the EC2 section or if I should put it after the templates.  I could fall either way since CloudFormation is an AWS thing and makes sense to follow the EC2 section, but on the other hand the use of scripts (for Linux and Windows) makes following immediately with the Custom Script Extension a good choice as well.

lorengordon:
I think we should add sections, `watchmaker in AWS` and `watchmaker in Azure` (at the same level as `watchmaker from the CLI`), and make the related sections sub-headings within those.

KevinPlus3:
Sounds good. Will work on that.

lorengordon:
@KevinPlus3 any more changes expected?

KevinPlus3:
Nope.  It is not as extensive as AWS section but I think Microsoft does a good job with documenting how to use it further.  If someone brings up expanding on it (possibly by adding a Resource Management Template like Eric experimented with), then maybe we can do that at a later time.

PR #467, opened 2017-12-15
add env command to the command blocks within the cfn lx templates

userhas404d:
Ensures the following aws envs are available in the appscript shell: 
- `AWS_DEFAULT_REGION`
- `AWS_CA_BUNDLE`
- `REQUESTS_CA_BUNDLE`

### DIFF
--- a/docs/files/templates/lx-instance/watchmaker-lx-instance.template
+++ b/docs/files/templates/lx-instance/watchmaker-lx-instance.template
@@ -641,7 +641,13 @@
                         {
                             "10-watchmaker-install" :
                             {
-                                "command" : "bash -xe /etc/cfn/scripts/watchmaker-install.sh"
+                                "command" : "bash -xe /etc/cfn/scripts/watchmaker-install.sh",
+                                "env" :
+                                {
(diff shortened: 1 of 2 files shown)


docs/files/templates/lx-instance/watchmaker-lx-instance.template
```diff
@@ -852,7 +858,14 @@
                                     { "Ref" : "AWS::Region" }, " &&",
                                     " chown root:root /etc/cfn/scripts/make-app &&",
                                     " chmod 700 /etc/cfn/scripts/make-app"
-                                ]]}
+                                ]]},
+                                "env" :
+                                {
+                                  "AWS_DEFAULT_REGION" : { "Ref" : "AWS::Region" },
```
lorengordon: Since we're passing in the region via the env, we could remove it from the aws s3 cp command itself.

lorengordon: And this block is missing in the autoscale edits...

userhas404d: fixed

Issue #468, opened 2017-12-20
Link references broken for certain heading anchor links in Watchmaker's Readthedocs website


eemperor:
Certain heading anchor links in html generated for [Watchmaker Readthedocs](https://watchmaker.readthedocs.io/en/stable/index.html) do not have the heading name but instead is given an arbitrary `id<#>` name.

https://watchmaker.readthedocs.io/en/stable/configuration.html#id1
https://watchmaker.readthedocs.io/en/stable/configuration.html#id2
https://watchmaker.readthedocs.io/en/stable/contributing.html#id1

This causes link references to the heading to not work properly.

lorengordon:
Closed by https://github.com/plus3it/watchmaker/pull/469

PR #469, opened 2017-12-20
Modified heading text to fix anchor link issue

eemperor:
Workaround for [#468](https://github.com/plus3it/watchmaker/issues/468) 

Problem appears to be due to a bug somewhere in the Markdown->reStructuredtext->HTML conversion chain (m2r, sphinx).  When the heading name and reference link name are the same and the reference link is placed **before** the heading, the heading anchor link is given an arbitrary `id<#>` name and not the heading name.

Making the heading name and reference link name different fixes the issue.  The other workaround is to place the reference link after the heading but this is not feasible because it would require a re-write of the document.


eemperor:
@lorengordon So true!

PR #491, opened 2018-01-10
Azure CSE PowerShell

Phydeauxman:
Python/Watchmaker install:
	- Changed the Python install to use "Start-Process" with -wait
	- Explicitly modify the session path variable to add Python
	  folders
	- Separated the download/install of Watchmaker from the other
	  Python modules
	- Set the download destination for the Python module to be the
	  $env:windir\Temp folder
	- Added code to delete all non-log files from the
	  $env:windir\Temp folder

### DIFF
--- a/docs/files/bootstrap/watchmaker-bootstrap.ps1
+++ b/docs/files/bootstrap/watchmaker-bootstrap.ps1
@@ -143,4 +143,9 @@ if( ${GitUrl} ) {
 Reset-EnvironmentVariables
 Write-Verbose "Reset the PATH environment for this shell"
 
+if ("$Env:TEMP".TrimEnd("\") -eq "${Env:windir}\System32\config\systemprofile\AppData\Local\Temp") {
+  $Env:TEMP, $Env:TMP = "${Env:windir}\Temp", "${Env:windir}\Temp"
+  Write-Verbose "Forced TEMP envs to ${Env:windir}\Temp"
+}
+
 Write-Verbose "${__ScriptName} complete!"


lorengordon:
Hmm, I think it makes the _example_ userdata too long and too complicated for the average user to understand. The changes are addressing environmental issues, not watchmaker issues. The documented userdata is intended only as an example. We purposefully keep it as simple as we can. Users are expected to manage their own userdata for their specific use cases.

So, I think I'm a negative on this one, as is. I can maybe be convinced otherwise, but it would need to be stripped down to bare essentials and the overly syntax-y bits would need to be removed.

docs/usage.md
```diff
@@ -132,10 +132,14 @@ $BootstrapFile = "${Env:Temp}\$(${BootstrapUrl}.split('/')[-1])"
...
 # Install python
-& "$BootstrapFile" -PythonUrl "$PythonUrl" -Verbose -ErrorAction Stop
+$params = "`"$BootstrapFile`" -PythonUrl `"$PythonUrl`" -Verbose -ErrorAction Stop"
+Start-Process powershell -Argument $params -NoNewWindow -Wait
+$env:Path = "$env:Path;$env:ProgramFiles\Python36\Scripts\;$env:ProgramFiles\Python36\"
 
 # Install watchmaker
-pip install --index-url="$PypiUrl" --upgrade pip setuptools watchmaker
+pip install --index-url="$PypiUrl" --upgrade pip setuptools boto3
+pip download --dest $env:windir\Temp watchmaker
```
lorengordon: Need to maintain the `--index-url="$PypiUrl"` in this command.

Phydeauxman: Reverted to the original pip line and retested.  Worked end to end so the only change that needs to be made is to install Python using Start-Process versus the Call Operator.

lorengordon:
Ok, we've been working on this out-of-band for a while and I wanted to summarize what we've found.

1. The problems with the userdata (custom/install script) occur when the execution context results in the usage of a particular TEMP path on Windows, `$env:windir\system32\config\systemprofile\AppData\Local\Temp`
2. We've found two issues around this TEMP path:
    * pip by default extracts the packages into the TEMP path, which results in MAXPATH errors for the bundled pshelp salt formula
    * apply_lgpo_delta writes and reads from the TEMP path and for some currently unknown reason this path causes apply_lgpo_delta to hang (as these particular files should not run into MAXPATH limits)
3. The problems are exposed when running on Azure with a Custom Script Extension because it seems the script is executed in the SYSTEM user context
    * TEMP starts out as `C:\temp`, but something in the python exe installer causes TEMP to be reset to the SYSTEM user's context `C:\system32\config\systemprofile\AppData\Local\Temp`
4. The problems **do not** occur in EC2 because the script is executed in the Administrator user's context, where TEMP is `C:\Users\Administrator\AppData\Local\Temp\`

We have a couple design goals in this area of the project, around the custom script, to keep in mind when coming up with a solution:

1. Keep the example custom script as simple and minimal as possible
2. Attempt to keep the custom script as similar as possible across platforms, windows and linux

Since windows requires more work to get a python application installed and running, we provide the watchmaker-bootstrap.ps1 script to  to get a windows system to the point where installing and using watchmaker is basically the same on windows as it is on linux. In support of that, we currently use the bootstrap to (also not ruling out other uses to address additional issues):

1. Install python (required for watchmaker)
2. Install git for windows (in a development environment)
3. Install certificates into the system cert store (in environments with custom CAs)
4. Manage necessary environment variables (such as refresh the PATH based on registry updates made by the installers)

As for potential solutions/workarounds for the issue at hand:

1. Use `Start-Process` to execute the bootstrap that installs python, and set the PATH in the custom script. This is the approach currently demonstrated [in this PR](https://github.com/plus3it/watchmaker/pull/491/files#diff-4be2078a125fe364f7198675d51bc688R137). This works because the custom script and the bootstrap script then run in different processes with different envs. So when the python installer in the bootstrap modifies the TEMP path, it doesn't not carry over to the parent custom script process.
2. (Needs testing) Update watchmaker-bootstrap.ps1 to manage the TEMP env. This approach would test for when the TEMP env is set to the SYSTEM user path and change it to the windir path. Something like:

    ```
    if ("$Env:TEMP" -eq "${Env:windir}\System32\config\systemprofile\AppData\Local\Temp")
    {
        $Env:TEMP, $Env:TMP = "${Env:windir}\Temp", "${Env:windir}\Temp"
    }
    ```

As it better supports the design goals, I would prefer option 2, if it does indeed work as I expect.

PR #493, opened 2018-01-18
Storage-handler for Fifth-generation Instance-types

ferricoxide:
Adds mapping-keys for "internal" and "external" device-names to be used in the `UserData` and `BlockDeviceMappings` sections, respectively. Legacy instance-types get `/dev/xvdf` assigned for both values. Selectable `m5` and `c5` instance-types get `/dev/xvdf` for "external" device-name and `/dev/nvme1n1` for "internal" device-name.

Mappings have been added to both the `docs/files/templates/lx-instance/watchmaker-lx-instance.template` and `docs/files/templates/lx-autoscle/watchmaker-lx-autoscale.template` files.

Tested against `m4.large` and `m5.large` instance-types.

### DIFF
--- a/docs/files/templates/lx-autoscale/watchmaker-lx-autoscale.template
+++ b/docs/files/templates/lx-autoscale/watchmaker-lx-autoscale.template
@@ -1,1199 +1,1293 @@
 {
-    "AWSTemplateFormatVersion" : "2010-09-09",
-    "Description" : "This template creates an Autoscaling Group and Launch Configuration that deploys Linux instances with Watchmaker, which applies the DISA STIG.",
-    "Parameters" :
-    {
-        "AmiId" :
-        {
-            "Description" : "ID of the AMI to launch",
-            "Type" : "String",
-            "AllowedPattern" : "^ami-[0-9a-z]{8}$|^ami-[0-9a-z]{17}$"
-        },
-        "AmiDistro" :
-        {
-            "Description" : "Linux distro of the AMI",
-            "Type" : "String",
-            "AllowedValues" :
-            [
-                "AmazonLinux",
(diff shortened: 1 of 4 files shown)


lorengordon:
Can we do this in a way where the diff is actually usable? Seems like there's maybe a lot of changes here unrelated to the feature being added...

lorengordon:
Ok, I think I figured out the approach by looking at the third and fourth commits. That part looks great. What's going on with the rest of the changes? The diff is way too big to figure it out on my phone...

ferricoxide:
I can re-do it... mostly the `diff` is the way it is because the json-formatter re-organized the sections (and subsections within each) to be in alphabetical order.

lorengordon:
Oh I see. If that's all no worries. We're about to move these into a new repo where we'll be running linter tests to enforce that type of thing anyway.

Maybe just take a look at the terraform files and param maps to update the modified parameter value.

ferricoxide:
Yeah: when I work on templates, I run them through python's json.tool module: it makes the file easier to read and puts things in consistent locations (across templates). It's basically reflex, at this point (i.e., wasn't considering the impact to the diff when I ran it).

With respect to:

> Maybe just take a look at the terraform files and param maps to update the modified parameter value.

Do you mean see if I can figure out the cognates for TF and add to to the TF templates, too?

lorengordon:
Pretty much the tf files just pass params to the cfn templates, so no real change or logic to consider. Just change the default value for the param/var and update the description.

lorengordon:
These templates have been moved to https://github.com/plus3it/terraform-aws-watchmaker. The work in this pr has now been incorporated in that repo.

PR #518, opened 2018-02-13
Sets name and body when deploying to GitHub releases

lorengordon:


.travis.yml
```diff
@@ -66,7 +66,8 @@ before_deploy:
     elif [ "$TRAVIS_BRANCH" = "master" ]
     then
       WAM_VERSION=$(grep "version = " $TRAVIS_BUILD_DIR/setup.cfg | sed "s/version = //")
-      export WAM_VERSION
+      WAM_BODY = "* [Watchmaker v$WAM_VERSION CHANGELOG](https://github.com/plus3it/watchmaker/blob/$WAM_VERSION/CHANGELOG.md)"
+      export WAM_VERSION WAM_BODY
```
KevinPlus3: You don't have export WAM_VERSION before setting the string for WAM_BODY?

lorengordon: Pretty sure it's the same shell at that point so the env is available without exporting it. The export is to ensure it is still available in the `deploy` block later. Also, the release is still published in draft mode for the moment, so we can manually correct it if this isn't quite right.

PR #532, opened 2018-02-22
Ensures downloads support tls11 and tls12 if they are available

lorengordon:


lorengordon:
Not currently. I don't know of a registry of python and git for windows releases, which we'd need to figure out first to get to what I _think_ you are proposing.

PR #613, opened 2018-05-30
Add standalone deployment to GitHub Rel and S3

### LINKED ISSUE #612: Add automated build/deploy of standalone versions for Windows/Linux

YakDriver:
Fixes #612


.travis.yml
```diff
@@ -1,110 +1,156 @@
...
+    after_failure:
+    - more .tox/log/* | cat
+    - more .tox/*/log/* | cat
+  - stage: deploy
+    if: branch in (master, develop)
+    python: 2.7
+    install: pip install m2r
+    script: echo "Deploying to PyPI ..."
+    before_deploy:
+    - if [ "$TRAVIS_BRANCH" = "develop" ]; then
```
lorengordon: I _think_ this needs to be written as a multiline string?

```
- |	
  if [ "$TRAVIS_BRANCH" = "develop" ]	
  then	
    sed -i -E "s/^(version = )(.*)/\1\2.$TRAVIS_BUILD_NUMBER/" $TRAVIS_BUILD_DIR/setup.cfg
  fi
```

lorengordon: Also, if we aren't going to use travis_set_build.py anymore, then please delete the file...

appveyor.yml
```diff
@@ -1,12 +1,18 @@
...
+image: Visual Studio 2013
 cache:
   - '%LOCALAPPDATA%\pip\Cache'
 matrix:
   fast_finish: true
 environment:
   global:
+    DEV_BUCKET: watchmaker-dev
+    RELEASE_BUCKET: watchmaker
+    SATS_TOKEN:
```
lorengordon: would prefer this to be set in the appveyor service, and removed from appveyor.yml

.travis.yml
```diff
@@ -1,110 +1,156 @@
...
+      skip_cleanup: true
+      skip_upload_docs: true
+      on:
+        tags: true 
+  - stage: deploy
+    if: branch in (master, develop)
+    python: 3.6
+    env:
+    - secure: HYd2fekdYkZts+j2hOXdQE6SZP7sJ8vBkSMovnGkZ4Clyx2jq0e7TbO3BYkM7JNm+1SEblKLTXWOc0uy7kzu1lZhWA15L4Gu5QrMUNiCcIvL9Hares7vPqVg8mduxBWUqXIHz+ZNQfHhysTWBmUJPyEq+fIkdqxeCY3YKLO+lb75r3BFqmz2Nt6/A0jfjKMPbN7RpbSxRtkejvLzk7Z0buUFdp9zCVCo4dinBnFFWtcGLqOJ0TA4Ko3x6P4qetui34ZV7eQj9Y9Zt+k3Nq7hHZbGZFqNEWvJQLHAF1Ohf3LHzibzMccBp9PycgBmgMNCLkUb2VvAkj7BcKur30QuY7VCGSr3C9z4Vb1NH+KTHS+ONkIy11SlvOQ0yceiTlnVobz11axaQzcLHH/vjph2ExjwdczjdS/M0NiT25+YQ9Gngc6qiPJ+fDx6czNRwiIQpvoiOCy7mQvBVudXp1Hp0Jr12ME3Ui6GTkRXjgXAMzs17u74GgwNscrsbg4ggs05gw5fdg61m3mer9YS2FW6PmMgS0Tnsu3jOZ4H4B3OO3l+RdrDOBxsXLfEw6xlNbX3tNF+KSUicNri4Oillh/6PkUeKSNAxvH0g5my0b0VYDZb5F7nu8tvSPjNdfxbVnEUiLISG+1YTbBwJCrX3bX1PFBg20m1UdO7rey0h++biNs=
```
lorengordon: would prefer this to be set in the travis service, and removed from .travis.yml

appveyor.yml
```diff
@@ -58,6 +68,53 @@ on_failure:
...
+  # upload & rename artifacts to appveyor's magic artifact depot in the sky (DITS)
+  - appveyor PushArtifact %GB_ENV_GEN_FILE_W_PATH% -FileName %GB_ENV_APP_VERSION%\%GB_ENV_GEN_FILE%
+  - appveyor PushArtifact %GB_ENV_APP_NAME%-%GB_ENV_APP_VERSION%-sha256-%GB_ENV_OPERATING_SYSTEM%-%GB_ENV_MACHINE_TYPE%.json -FileName %GB_ENV_APP_VERSION%\%GB_ENV_APP_NAME%-%GB_ENV_APP_VERSION%-sha256-%GB_ENV_OPERATING_SYSTEM%-%GB_ENV_MACHINE_TYPE%.json
+  - appveyor PushArtifact %GB_ENV_GEN_FILE_W_PATH% -FileName latest\%GB_ENV_APP_NAME%-latest-standalone-%GB_ENV_OPERATING_SYSTEM%-%GB_ENV_MACHINE_TYPE%.exe
+  - appveyor PushArtifact %GB_ENV_APP_NAME%-%GB_ENV_APP_VERSION%-sha256-%GB_ENV_OPERATING_SYSTEM%-%GB_ENV_MACHINE_TYPE%.json -FileName latest\%GB_ENV_APP_NAME%-latest-sha256-%GB_ENV_OPERATING_SYSTEM%-%GB_ENV_MACHINE_TYPE%.json
+  - SET
+  - ps: ls
+  - ps: >-
+      if($env:APPVEYOR_REPO_BRANCH -eq "master") {
+        satsuki --verbose
```
lorengordon: does appveyor not have an equivalent of travis's "script" deploy provider?

appveyor.yml
```diff
@@ -58,6 +68,53 @@ on_failure:
...
+  - ps: >-
+      if($env:APPVEYOR_REPO_BRANCH -eq "master") {
+        $env:S3_BUCKET=$env:RELEASE_BUCKET
+      } else {
+        $env:S3_BUCKET=$env:DEV_BUCKET
+      }
+  - set S3_KEYFIX=releases
+  - gravitybee-environs.bat
+  # upload & rename artifacts to appveyor's magic artifact depot in the sky (DITS)
+  - appveyor PushArtifact %GB_ENV_GEN_FILE_W_PATH% -FileName %GB_ENV_APP_VERSION%\%GB_ENV_GEN_FILE%
```
lorengordon: this block, where directory names and file names/paths are being contructed from numerous assorted ENVs, is really hard to read and comprehend what is going on... maybe set some well-named ENVs instead of stringing together all these different values? or rethink what gravitybee is outputting so this is easier to handle?

appveyor.yml
```diff
@@ -58,6 +68,53 @@ on_failure:
...
   - path: dist\*
 
+# scripts to run before deployment
+before_deploy:
+  # install gravitybee/generate the standalone executable
+  - mkdir C:\venv
+  - C:\Python36-x64\Scripts\virtualenv C:\venv
+  - C:\venv\Scripts\activate
+  - python --version
+  - pip install gravitybee satsuki
```
lorengordon: i think we should version-pin these build/deploy dependencies in a requirements/deploy.txt. i don't like being surprised when ci/cd dependencies change and break our builds... the install then would be `pip install -r requirements/deploy.txt`.

appveyor.yml
```diff
@@ -1,12 +1,18 @@
 version: '{branch}-{build}'
-build: off
+build: false
+image: Visual Studio 2013
```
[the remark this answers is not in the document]
YakDriver: The default build platform is VS 2015. You can pick from 13, 15, or 17 and 13 seems faster.

appveyor.yml
```diff
@@ -37,19 +43,23 @@ environment:
...
+- ps: echo $env:TOXENV
+- ps: echo "Python $env:PYTHON_VERSION (${env:PYTHON_ARCH}bit) from $env:PYTHON_HOME"
+- ps: ls C:\Python*
 install:
-  - git submodule update --init --recursive
-  - '%PYTHON_HOME%\Scripts\pip install --upgrade tox setuptools virtualenv wheel'
-  - '%PYTHON_HOME%\Scripts\virtualenv --version'
-  - '%PYTHON_HOME%\Scripts\pip --version'
-  - '%PYTHON_HOME%\Scripts\tox --version'
+- ps: $env:DO_DEPLOY="no"
```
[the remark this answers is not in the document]
YakDriver: Yes there is. Whether `before_deploy` is reached, depends on if `deploy` will be reached, not vice versa. Before AppVeyor reaches `before_deploy` or `deploy`, it determines if the conditions are met to `deploy`. If the conditions are met to `deploy`, then it runs `before_deploy`, followed by `deploy`. So, this can be in any of the earlier stages but not in `before_deploy`.

appveyor.yml
YakDriver: Nope. It is listed in the docs but is not a link like the other providers. I tried using it to test but it says it is not recognized.

appveyor.yml
```diff
@@ -58,6 +68,53 @@ on_failure:
...
+  - provider: S3
+    access_key_id:
+      secure: IgUARv6O1DqRDRKQ3rjZEuDdPGuzz/QCzjkwdWRqx5o=
+    secret_access_key:
+      secure: RiIvGJ442UiwnoeZ53D5mpLKXwn/e5PSEYxZR/FIL4cFIRkbnZTzGMiuMoEbXRcE
+    bucket: $(S3_BUCKET)
+    folder: releases
+    #artifact: # not needed since by default it will grab everything added to DITS
+    on:
+      do_deploy: yes
```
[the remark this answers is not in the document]
lorengordon: I think the workaround adds more code than a second deploy provider with a different condition, especially if the access/secret keys are set as envs in the service rather than the file... 😆 

It's declarative vs imperative styles. For these configs, declarative is much easier to read.

.travis.yml
YakDriver: I didn't purposely use a YAML formatter but when I used Travis's own tool to add something (the S3 stuff where I discovered undocumented attributes), it did some interesting things like remove the pipes from multiline things and take the spaces out from lists. I figured Travis was giving it as some sort of recommendation so I left it. Happy to put it back though.

appveyor.yml
YakDriver: Yeah, I'm attacking the same problem two different ways from two different iterations. It's confusing. I'll clean it up.

appveyor.yml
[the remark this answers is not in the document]
YakDriver: I'm non-denominational when it comes to declarative vs. imperative.

.travis.yml
```diff
@@ -53,58 +47,95 @@ notifications:
...
+        - export SATS_REL_NAME="Release v$GB_ENV_APP_VERSION"
+      deploy:
+        - provider: script
+          # deploy to github releases
+          script: satsuki --verbose
+          skip_cleanup: true
+          on:
+            branch: master
+        - provider: s3
+          access_key_id: $AWS_ACCESS_KEY
```
lorengordon: ought to be able to remove the keys `access_key_id` and `secret_access_key`, since the SDK will read from the ENVs by default

.travis.yml
```diff
@@ -53,58 +47,95 @@ notifications:
...
+          secret_access_key: $AWS_SECRET_ACCESS_KEY
+          bucket: $RELEASE_BUCKET
+          upload-dir: $S3_KEYFIX
+          local-dir: $GB_ENV_STAGING_DIR
+          skip_cleanup: true
+          on:
+            branch: master
```
lorengordon: i think this condition should be `tags: true`, so it deploys to s3 at the same time it deploys to pypi

.travis.yml
```diff
@@ -53,58 +47,95 @@ notifications:
...
-      branch: master
-      repo: plus3it/watchmaker
-      condition: '"$TOXENV" == *"py27"*'
+jobs:
+  include:
+    - stage: check
+      python: 3.6
+      env:
+        - TOXENV=check
+      install:
```
lorengordon: i suspect we could remove the `install`, `script`, and `after_failure` sections, since they are the same as the settings at the top level, so they should get "inherited" when the stage section is merged into the job...

.travis.yml
```diff
@@ -53,58 +47,95 @@ notifications:
...
+      script:
+        - tox -v
+      after_failure:
+        - more .tox/log/* | cat
+        - more .tox/*/log/* | cat
+    - stage: docs
+      python: 3.6
+      env:
+        - TOXENV=docs
+      install:
```
lorengordon: same with this job

appveyor.yml
```diff
@@ -58,6 +58,44 @@ on_failure:
...
+  - provider: Environment
+    name: wam-s3-dev
+    on:
+      branch: develop
+      toxenv: py36,codecov
+      python_arch: 64
+  - provider: Environment
+    name: wam-s3-releases
+    on:
+      branch: master
```
lorengordon: ok, yeah, i said this should happen on `branch: master`, but think it should actually be `appveyor_repo_tag: true`...

appveyor.yml
```diff
@@ -58,6 +58,44 @@ on_failure:
...
+  - ps: ls
+  # deploy to github releases
+  - ps: >-
+      if($env:APPVEYOR_REPO_BRANCH -eq "master") {
+        satsuki --verbose
+      }
+# deploy to s3
+deploy:
+  - provider: Environment
+    name: wam-s3-dev
```
lorengordon: where is it defined what this actually does?

YakDriver: At AppVeyor website, at the top you'll see "ENVIRONMENTS" to right of "PROJECTS." Click that and you'll see the environments listed by name. Click on either one, then "SETTINGS" and you'll see the details (e.g., access key ID, bucket name, folder).

lorengordon: Oh, that's at the global appveyor level, rather than the project level... is there some reason not to use the project settings? https://ci.appveyor.com/project/plus3it/watchmaker/settings/deployment

YakDriver: You can't use the fancy environment feature in project settings.

YakDriver: Never mind. You can do it all in project settings...

lorengordon: looks like it could also be setup more similarly to travis:
```
deploy:
- provider: S3
  access_key_id: foo
  secret_access_key:
    secure: fhffELWjrOuP7+fHzE4lRQ==
  region: us-east-1
  bucket: baz
  on:
    branch: develop
```

YakDriver: It's actually cleaner in the project settings because you can put the conditions there as well.

YakDriver:
What I've learned about appveyor deployment:

1. You can setup ENVIRONMENTS but those are global to the account, not per project
2. You can setup DEPLOYMENT settings in a project but there's not evidence of the deployment in the appveyor.yml.
3. If you include both `deploy_script` and `deploy`, it will ignore the `deploy`.

appveyor.yml
YakDriver: Actually this is not entirely true. AppVeyor has a stage/step called `deploy_script:` but it seems to be mutually exclusive with the `deploy:` stage/step. (More research needed.)

lorengordon:
Alright, looks good to me. Merge if you're ready.

PR #617, opened 2018-06-07
Added documentation branding

moskey71:
Added P3 logo and linked the image to our web site


docs/api.md
```diff
@@ -1,3 +1,10 @@
+<p align="right">
```
lorengordon: looks like these image directives will all need to be written in restructured text, or they won't render correctly... for example:

```
.. image:: images/cropped-plus3it-logo-cmyk.png
   :width: 140px
   :alt: Powered by Plus3 IT Systems
   :align: right
   :target: https://plus3it.com
```

http://docutils.sourceforge.net/docs/ref/rst/directives.html#image

moskey71: I've updated the image code to use RST.  The pages rendered correctly after generating them using sphinx on my local system.  Thanks for the reference

docs/api.md
```diff
@@ -1,3 +1,13 @@
...
+<b>Powered by:</b>
```
lorengordon: This "Powered by" looks pretty cheesy to me, and a bit out of place, spacing-wise. Can we remove that and just use the image?

image

If we want more, maybe drop a new heading at the bottom of the Index page, Maintained By, or Provided By, or somesuch, with a short paragraph about plus3it and some links to p3 resources (github, home page, etc).

lorengordon: But I'm fine with just the image for now, and we can let that stew before doing anything more.

moskey71: @Briankwatson was fine with removing the text.  Code has been updated to reflect the change

Issue #619, opened 2018-06-09
Use sha as build number in dev pkgs, instead of CI build number


lorengordon:
We've been pushing dev pkgs to testpypi using travis for quite a while, but we recently also added standalone binaries that are pushed to an S3 bucket. Those packages need to be created on a like platform (i.e. a Windows pkg needs to be created on a Windows system). So now we have appveyor building Windows standalone pkgs and travis building Linux standalone pkgs.

The build version ends up the same, despite being built in two different places, because we aren't actually setting the build version in the CI jobs for these standalone pkgs.

* https://github.com/plus3it/watchmaker/blob/develop/.travis.yml#L88
* https://github.com/plus3it/watchmaker/blob/develop/appveyor.yml#L61

This leads to a discrepancy between the testpypi dev version and the standalone dev version.

We can align the two by using the sha of the built commit as the build number, for all dev deploys, i.e. `x.x.x.dev<sha>`.

YakDriver:
Fixed in PR #621

PR #621, opened 2018-06-12
Fix version for dev branch

### LINKED ISSUE #618: Ensure dev packages contain "dev" version
We update the version with the "dev" indicator after publishing a release, but we sometimes forget to do so in a timely manner. Other PRs get merged, before the "dev" indicator is in place. This doesn't really impact any actual release, but it does lead to misnamed/misordered packages in testpypi:

* https://test.pypi.org/project/watchmaker/#history

We ought to be able to avoid this by changing how we add the "build" to the version, in .travis.yml. I.e. also inject the "dev" indicator at the same time.

* https://github.com/plus3it/watchmaker/blob/develop/.travis.yml#L68

YakDriver:
Fixes #618 and #619

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -65,6 +65,12 @@ before_deploy:
   - C:\venv\Scripts\activate
   - python --version
   - pip install -r requirements/deploy.txt
+  - ps: >-  
+      if($env:APPVEYOR_REPO_BRANCH -eq "develop") {
+        # whether setup.cfg version has "dev" or not, it is added but not duplicated
+        $sha = ($env:APPVEYOR_REPO_COMMIT).Substring(0,7)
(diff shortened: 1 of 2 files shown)


appveyor.yml
```diff
@@ -65,6 +65,12 @@ before_deploy:
   - C:\venv\Scripts\activate
   - python --version
   - pip install -r requirements/deploy.txt
+  - ps: >-
```
lorengordon: Were you able to test this block? Or drop it in a yaml formatter and then check the resulting string? I'm suspecting we may need semicolons after each line to separate the commands properly...

YakDriver: In S3, watchmaker-dev/releases, the latest build from my fork is 0.9.7.dev73727f6. The AppVeyor and Travis files are living together in perfect harmony.

PR #623, opened 2018-06-13
Switch from matrix to linear testing

### LINKED ISSUE #620: Speed-up builds by not using AppVeyor matrix feature
The matrix feature creates separate jobs for each platform, which are then queued. Queuing time on AppVeyor can be significant for a single job and is multiplied with multiple non-parallel jobs. In order to reduce queuing time, and thus the time for AppVeyor to complete an overall build, the number of jobs has to be reduced. 

It is possible to manually matrix tasks, set environment variables appropriately, and thus to run all the tox tests and build tasks within a single job.

Downsides to this manual matrix, or "linear" approach, include less clean code and potential for error with manually duplicating code blocks. These can be reduced through tidy indentation and spacing of code. The main upside is performance. Tests have shown that the linear approach can perform the same build tasks in 41% of the time of the matrix approach.

YakDriver:
Fixes #620

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -3,52 +3,91 @@ build: off
 image: Visual Studio 2013
 cache:
   - '%LOCALAPPDATA%\pip\Cache'
-matrix:
-  fast_finish: true
 environment:
   global:
     WITH_ENV: 'cmd /E:ON /V:ON /C .\ci\run_with_env.cmd'
-  matrix:
-    - TOXENV: check
-      PYTHON_HOME: C:\Python36-x64
-      PYTHON_VERSION: 3.6
-      PYTHON_ARCH: 64
 
-    - TOXENV: docs
-      PYTHON_HOME: C:\Python36-x64
-      PYTHON_VERSION: 3.6
-      PYTHON_ARCH: 64
-
-    - TOXENV: '2.7,codecov'
-      PYTHON_HOME: C:\Python27
-      PYTHON_VERSION: 2.7
-      PYTHON_ARCH: 32
-
-    - TOXENV: '2.7,codecov'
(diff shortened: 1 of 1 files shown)


appveyor.yml
```diff
@@ -3,52 +3,68 @@ build: off
...
   - '%PYTHON_HOME%\Scripts\tox --version'
-test_script:
+  - '%PYTHON_HOME%\Scripts\pip --version'
+
+  # -----
+  - set TOXENV=docs
+  - set PYTHON_HOME=C:\Python36-x64
+  - set PYTHON_VERSION=3.6
+  - set PYTHON_ARCH=64
+  - '%PYTHON_HOME%\Scripts\pip install --upgrade tox setuptools virtualenv wheel'
```
lorengordon: you should only need to install these setup pkgs once per python_home... i get this makes the sections very similar, so maybe group sections by python_home?

appveyor.yml
```diff
@@ -3,52 +3,68 @@ build: off
...
+  - set PYTHON_ARCH=64
   - '%PYTHON_HOME%\Scripts\pip install --upgrade tox setuptools virtualenv wheel'
+  - '%WITH_ENV% %PYTHON_HOME%\Scripts\tox -v'
+
+  # check versions
   - '%PYTHON_HOME%\Scripts\virtualenv --version'
-  - '%PYTHON_HOME%\Scripts\pip --version'
   - '%PYTHON_HOME%\Scripts\tox --version'
-test_script:
+  - '%PYTHON_HOME%\Scripts\pip --version'
```
lorengordon: i think generally we do the version checks _before_ running the test script (tox, in this case)... perhaps also grab these once per python_home...

YakDriver:
This is cleaner.

PR #632, opened 2018-06-22
Allow system CA certs to be used

YakDriver:


src/watchmaker/utils/__init__.py
```diff
@@ -45,4 +46,6 @@ def basename_from_uri(uri):
 @backoff.on_exception(backoff.expo, urllib.error.URLError, max_tries=5)
 def urlopen_retry(uri):
     """Retry urlopen on exception."""
-    return urllib.request.urlopen(uri)
+    # "trust the system’s default CA certificates"
```
lorengordon: nit...that single quote looks like a special unicode char...mind changing it to plain jane single quote?

YakDriver:
Trouble with copy-paste from docs. Single quote fixed.

PR #640, opened 2018-07-06
Build standalones with improved compatibility.

YakDriver:
Probably 75-100 commits squashed into this rather small PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -90,11 +90,19 @@ jobs:
       if: branch in (master, develop)
       python: 3.6
       install:
-        - pip install -r requirements/deploy.txt
-        - pip install --editable . # install wam, as needed by GB
+        - sudo docker pull ${DOCKER_SLUG}:latest
       script:
-        # create standalone application
-        - gravitybee --src-dir src --sha file --with-latest --extra-data static --verbose
+        - echo "DOCKER_INSTANCE_NAME:${DOCKER_INSTANCE_NAME}"
+        - echo "DOCKER_SLUG:${DOCKER_SLUG}"
+        - |
+          sudo docker run --detach --privileged \
+            --volume="${TRAVIS_BUILD_DIR}":"${TRAVIS_BUILD_DIR}" \
+            --workdir "${TRAVIS_BUILD_DIR}" \
+            --name ${DOCKER_INSTANCE_NAME} \
+            ${DOCKER_SLUG}:latest \
+            init
+        - sudo docker exec ${DOCKER_INSTANCE_NAME} chmod +x ci/deploy.sh
+        - sudo docker exec ${DOCKER_INSTANCE_NAME} ci/deploy.sh
       before_deploy:
         # prepare deployment envs
         - source .gravitybee/gravitybee-environs.sh
@@ -102,6 +110,8 @@ jobs:
         - export SATS_BODY="* [$GB_ENV_APP_NAME v$GB_ENV_APP_VERSION CHANGELOG](https://github.com/plus3it/$GB_ENV_APP_NAME/blob/$GB_ENV_APP_VERSION/CHANGELOG.md)"
         - export SATS_REL_NAME="Release v$GB_ENV_APP_VERSION"
         - echo Tag name is $SATS_TAG
+      after_script:
+        - sudo docker stop ${DOCKER_INSTANCE_NAME}
       deploy:
         - provider: script
           # deploy to github releases
--- a/ci/deploy.sh
+++ b/ci/deploy.sh
@@ -1,0 +1,16 @@
+#!/bin/bash
+
+export VIRTUAL_ENV_DIR=venv
+
+virtualenv $VIRTUAL_ENV_DIR
+source $VIRTUAL_ENV_DIR/bin/activate
+
+echo "-----------------------------------------------------------------------"
+python --version
+pip --version
+echo "-----------------------------------------------------------------------"
+pip install -r requirements/deploy.txt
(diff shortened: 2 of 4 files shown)


.travis.yml
```diff
@@ -90,18 +90,26 @@ jobs:
       if: branch in (master, develop)
       python: 3.6
       install:
-        - pip install -r requirements/deploy.txt
-        - pip install --editable . # install wam, as needed by GB
+        - sudo docker pull ${DOCKER_USER}/${DOCKER_REPO}
       script:
-        # create standalone application
-        - gravitybee --src-dir src --sha file --with-latest --extra-data static --verbose
+        - |
```
lorengordon: I recognize the work involved in getting here. But. This seems like it should be fine as individual lines, rather than a multiline string... The problem with this approach is that only a failure in the last line will actually cause a failure in the job, making failures somewhat harder to troubleshoot.

.travis.yml
```diff
@@ -90,18 +90,26 @@ jobs:
       if: branch in (master, develop)
       python: 3.6
       install:
-        - pip install -r requirements/deploy.txt
-        - pip install --editable . # install wam, as needed by GB
+        - sudo docker pull ${DOCKER_USER}/${DOCKER_REPO}
```
lorengordon: What do you think of just DOCKER_IMAGE or DOCKER_SLUG, rather than two separate variables?

.travis.yml
```diff
@@ -90,18 +90,26 @@ jobs:
...
       script:
-        # create standalone application
-        - gravitybee --src-dir src --sha file --with-latest --extra-data static --verbose
+        - |
+          sudo docker run --detach --privileged \
+            --volume="${TRAVIS_BUILD_DIR}":"${TRAVIS_BUILD_DIR}" \
+            --workdir "${TRAVIS_BUILD_DIR}" \
+            --name centos-6 \
```
lorengordon: Can the name be the env of the docker image, rather than hard-coded?

src/watchmaker/managers/base.py
```diff
@@ -174,10 +174,37 @@ def call_process(self, cmd, log_pipe='all', raise_error=True):
...
 
         self.log.debug('Command: %s', ' '.join(cmd))
+
+        # If running as a standalone, PyInstaller will have modified the
+        # LD_LIBRARY_PATH to point to standalone libraries. If there were a
+        # value at runtime, PyInstaller will create LD_LIBRARY_PATH_ORIG. In
+        # order for salt to run correctly, LD_LIBRARY_PATH has to be fixed.
+        kwargs = {}
+        env = dict(os.environ)
+        lp_key = 'LD_LIBRARY_PATH'
```
lorengordon: Very strong preference for spelling out variable names. What is `lp`?

.travis.yml
```diff
@@ -90,18 +90,28 @@ jobs:
...
+        - echo "DOCKER_SLUG:${DOCKER_SLUG}"
+        - |
+          sudo docker run --detach --privileged \
+            --volume="${TRAVIS_BUILD_DIR}":"${TRAVIS_BUILD_DIR}" \
+            --workdir "${TRAVIS_BUILD_DIR}" \
+            --name ${DOCKER_INSTANCE_NAME} \
+            ${DOCKER_SLUG}:latest \
+            init
+        - sudo docker exec ${DOCKER_INSTANCE_NAME} chmod +x ci/deploy.sh
+        - sudo docker exec ${DOCKER_INSTANCE_NAME} ci/deploy.sh
```
YakDriver: Figured out that in YAML the line-continuation backslash doesn't work well unless you're in multi-line mode with the pipe. Guess that makes sense but forgot it or never came across it. It kept failing with `docker: invalid reference format` errors without being in multi-line. Now it's one command per line, as you requested, but one of the commands takes several lines.

YakDriver:
This should be ready for review.

ci/deploy.sh
```diff
@@ -0,0 +1,16 @@
...
+
+echo "-----------------------------------------------------------------------"
+python --version
+pip --version
+echo "-----------------------------------------------------------------------"
+pip install -r requirements/deploy.txt
+pip install --editable .
+
+# creates standalone 
+gravitybee --src-dir src --sha file --with-latest --extra-data static --verbose
```
lorengordon: Newline please at the end of this file to make the git/github diff happy...

src/watchmaker/utils/__init__.py
```diff
@@ -46,6 +46,15 @@ def basename_from_uri(uri):
 @backoff.on_exception(backoff.expo, urllib.error.URLError, max_tries=5)
 def urlopen_retry(uri):
     """Retry urlopen on exception."""
-    # "trust the system's default CA certificates"
-    ctx = ssl.create_default_context(ssl.Purpose.CLIENT_AUTH)
-    return urllib.request.urlopen(uri, context=ctx)
+    kwargs = {}
```
lorengordon: Any idea why this is showing up here? This work was merged in https://github.com/plus3it/watchmaker/pull/633/... maybe a bad commit/amend or something?

lorengordon: Comparing the diff of this content and the prior PR and I see zero difference...

YakDriver: Hmm... I must have messed up somewhere. I had so many commits that git got all messed up. It listed all the commits in the reverse order and such nonsense.

lorengordon:
Looks good now, merge if you're done.

PR #646, opened 2018-07-18
Use separate script for Docker build

YakDriver:
Splitting the build functionality into a separate script allows Terrafirm and Travis to use the same script for building standalones, minimizing redundancy.


ci/prep_docker.sh
```diff
@@ -0,0 +1,68 @@
+#!/bin/bash
+
```
lorengordon: Bash won't exit on error by default. Add `set -xe -o pipefail` to harden the script, or call the script with `bash -xe -o pipefail`...

YakDriver: I made this change, which of course caused a problem (it's not a problem if old Docker packages can't be removed since they may not be installed). Fixed that a couple other hardening issues (quoting subshell, variables).

PR #656, opened 2018-08-06
Fix code blocks to display correctly, avoid sphinx warnings

YakDriver:
Sphinx threw some warnings not liking powershell code blocks.

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -15,7 +15,7 @@ to use the default configuration), using Watchmaker as a CLI utility is as
 simple as executing `watchmaker`. Below is the output of `watchmaker --help`,
 showing the CLI options.
 
-```shell
+```console
 # watchmaker --help
 Usage: watchmaker [OPTIONS]
 
@@ -79,13 +79,13 @@ utility.
 For example, on Linux, you can view the CLI options (shown above) using
 the same flag.
 
-```shell
+```console
 # ./watchmaker --help
 ```
 
 From Windows, similarly, execute Watchmaker by running it from the command line:
 
-```powershell
+```ps1con
 PS C:\wam> watchmaker.exe --help
 ```
 
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -94,7 +94,7 @@ Linux.
     *   From PowerShell, the Windows package can be downloaded
         as follows:
 
-        ```powershell
+        ```ps1con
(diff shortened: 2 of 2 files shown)


docs/usage.md
```diff
@@ -16,7 +16,7 @@ simple as executing `watchmaker`. Below is the output of `watchmaker --help`,
 showing the CLI options.
 
 ```shell
-# watchmaker --help
+$ watchmaker --help
```
lorengordon: Rationale for changing this? watchmaker needs to execute as root or it won't have perms to do much of anything...

YakDriver: The shell code block greys it out as a comment. But, now it's confusing for a whole new reason. Maybe we just get rid of both?

lorengordon: I'd say if changing to `$` then add `sudo` also. Otherwise, this change and the one below it are not really related to the sphinx warnings so just remove them from this pr. Can readdress later if we think of a better way.

YakDriver:
Pygments documentation yielded two handy code-block types: `ps1con` and `console` for the type of blocks we're trying to handle.

docs/conf.py
```diff
@@ -73,7 +73,7 @@
 
 # General information about the project.
 project = u'Watchmaker'
-copyright = u'2016, Plus3 IT Systems'  # noqa: A001
```
lorengordon: Copyright matches the LICENSE. I haven't seen anything convincing to indicate it's really necessary to update the year.....

lorengordon:
Brilliant, good find. I know I've looked for those tags before and came up empty!

docs/conf.py
YakDriver: Four points: 1. legally, everything you reduce to tangible medium is protected by copyright regardless, even if you have no copyright notice. 2. having an up-to-date copyright year does provide an educational/legal advantage in that it shows you actively protect you intellectual property. people in cyberspace may erroneously think you've abandoned your work and they can help themselves (they are legally and technically wrong). 3. having the date up-to-date shows that the project is fresh, still being maintained, that the maintainers care, and 4. if you want to show the longstanding nature of a project, it is perfectly acceptable to put a range (e.g., (c) Plus3 IT Systems 2009-2018).

lorengordon: Understood, we can consider it separately. I'd have a concern about maintaining the year range across all our projects... Absent a solution for that, I'd be more inclined to just removing the year.

PR #657, opened 2018-08-08
Releases watchmaker 0.10.0

lorengordon:


eemperor:
Any concerns with the codecov test fails?

YakDriver:
The travis codecov tests all passed fine. IMHO these patch/project-level codecov percentages are 1) aspirational not necessarily show stoppers, and 2) difficult to achieve where a test harness is not available (i.e., `travis.yml` vs. python code).

YakDriver:
go codebuild go

lorengordon:
The codecov integration is purely informational. We don't have much in the way of unit tests, which drive coverage. Since we don't have the unit tests anyway, we haven't bothered figuring out what codecov is trying to tell us or managing the codecov config to better reflect what we care about.

eemperor:
Thanks for the explanations.  I have only done a cursory look at covecov so I wasn't what the errors really meant.

Issue #658, opened 2018-08-08
Deploy release to s3 and GitHub releases broken


YakDriver:
Need build requirement (gravitybee) within the docker build container but the deploy requirement (satsuki) within the Travis container. Since satsuki is only installed within the docker build container, it is not available in Travis. This has cascading effect of never creating the release/tag so that deploys to pypi, s3 release bucket, and github releases all never happen.

YakDriver:
Fixed in #660

PR #660, opened 2018-08-08
Fix issue with deploy/build requirements

YakDriver:


requirements/build.txt
```diff
@@ -0,0 +1 @@
+gravitybee==0.1.20
```
YakDriver: All these changes took about 10 minutes except for this one.

YakDriver:
Fixes #658

PR #711, opened 2018-10-29
Makes --admin-groups parameters lowercase on Linux systems

eemperor:
This change will convert `--admin-groups` parameters to lowercase on Linux systems.


eemperor:
go codebuild go

eemperor:
go codebuild go

eemperor:
go codebuild go

src/watchmaker/__init__.py
```diff
@@ -224,6 +224,10 @@ def __init__(self, arguments):
 
         self.log.debug('System Parameters: %s', self.system_params)
 
+        # Make 'admin_groups' lowercase on Linux systems
```
lorengordon: I think it would be better to handle this logic in `workers/salt.py`, since this argument is specific to that worker.

What we can do is delete this line:

* https://github.com/plus3it/watchmaker/blob/9805a27f684ab43e1bd7fd03799dab4c9d52f727/src/watchmaker/workers/salt.py#L96

And insert the same line into the SaltLinux and SaltWindows init functions.

* https://github.com/plus3it/watchmaker/blob/9805a27f684ab43e1bd7fd03799dab4c9d52f727/src/watchmaker/workers/salt.py#L520
* https://github.com/plus3it/watchmaker/blob/9805a27f684ab43e1bd7fd03799dab4c9d52f727/src/watchmaker/workers/salt.py#L669

And only in the SaltLinux init, force the value to lowercase.

eemperor:
go codebuild go

eemperor:
go codebuild go

eemperor:
@lorengordon 

I made your suggested changes.  It worked fine but `pylint` complained about `admin_groups` missing in the parent class.  Seems `pylint` is not able to resolve the value in the inhiriting classes.  Looks like a potential bug (see [pylin issue #2425](https://github.com/PyCQA/pylint/issues/2425)).

I added `admin_groups` to the initialization which made `pylint` happy.  Still workes fine.  Any concerns?

I'll work on the additional handling of replacing spaces with `^` and `--user-groups`.

src/watchmaker/workers/salt.py
```diff
@@ -669,6 +674,7 @@ def __init__(self, *args, **kwargs):
         # Pop arguments used by SaltWindows
         self.installer_url = kwargs.pop('installer_url', None) or ''
         self.ash_role = kwargs.pop('ash_role', None) or ''
+        self.admin_groups = kwargs.pop('admin_groups', None) or ''
```
lorengordon: This ought to be moved after the `super(...)`, so that the SaltBase init does not set the value back to `None`.

src/watchmaker/workers/salt.py
```diff
@@ -523,6 +523,11 @@ def __init__(self, *args, **kwargs):
             kwargs.pop('bootstrap_source', None) or ''
         self.git_repo = kwargs.pop('git_repo', None) or ''
         self.salt_version = kwargs.pop('salt_version', None) or ''
+        self.admin_groups = kwargs.pop('admin_groups', None) or ''
+
+        # Make admin_groups lowercase
+        if self.admin_groups is not None:
```
lorengordon: I don't think this test is necessary, since the way it is set/popped above ensures it is always a string.

src/watchmaker/workers/salt.py
```diff
@@ -93,12 +93,12 @@ def __init__(self, *args, **kwargs):
         self.salt_debug_log = kwargs.pop('salt_debug_log', None) or ''
         self.salt_content = kwargs.pop('salt_content', None) or ''
         self.ou_path = kwargs.pop('ou_path', None) or ''
-        self.admin_groups = kwargs.pop('admin_groups', None) or ''
         self.admin_users = kwargs.pop('admin_users', None) or ''
         self.salt_states = kwargs.pop('salt_states', None) or ''
         self.exclude_states = kwargs.pop('exclude_states', None) or ''
 
         # Init attributes used by SaltBase, overridden by inheriting classes
+        self.admin_groups = None
```
lorengordon: Quite right, removing it entirely from SaltBase was wrong.

lorengordon:
I just realized I was thinking about this wrong. We can go back to popping `admin_groups` in the `SaltBase` init. Then, in `SaltLinux`, after calling `super()`, just modify it with `self.admin_groups = self.admin_groups.lower()` since the super() will set it to the initial value.

src/watchmaker/workers/salt.py
eemperor: Right.  It'll be a string with values or just an empty string.

eemperor:
I see what you mean.  It should be present after inhiriting it from `SaltBase`.  I'll rework it and see.

eemperor:
go codebuild go

eemperor:
I'm probably overdueing the test but I believe so!

PR #733, opened 2018-12-03
Ensure environment selection is valid

confusdcodr:
Previously, end user could choose any type of environment
End user will now be limited to choosing from '', 'dev', 'test', 'prod'


confusdcodr:
I will write some tests and documentation. I previously tested this simply by trying to run the tool on a CentOS7 vagrant box

src/watchmaker/__init__.py
```diff
@@ -251,6 +251,25 @@ def __init__(self, arguments):
 
         self.config = self._get_config()
 
+        # Ensure a valid environment type was selected
+        if 'salt' in self.config:
```
lorengordon: A couple of things here... 

1. I think we can flatten these nested if statements by using the dict.get method with a default value. I.e...

```
salt = self.config.get('salt', {'config': {}})['config']
valid_environments = [str(x).lower() for x in salt.get('valid_environments', [])]
env = str(salt.get('environment', '')).lower()
if env not in valid_environments:
    msg = (
        'Selected environment ({}) is not one of the valid'
        ' environment types: {}'.format(env, valid_environments)
    )
    self.log.critical(msg)
    raise InvalidValue(msg)
```

2. However, all these checks are an indication we're doing this in the wrong place. And I'm a little wary of putting worker-specific logic (salt) in the watchmaker `__init__` anyway... We can do it this way for now, but if you want to take a shot at putting the logic in the worker, that would be ideal. Right now the contract with the workers executes only a `install()` method. But, we can extend the contract to run a `before_install()` method where we can do this type of config validation.

That will require some teasing apart of the `worker_cadence` methods in `managers/workers.py`, as well as adding the new `before_install()` method to the base classes in `workers/yum.py` and `workers/salt.py`.

src/watchmaker/static/config.yaml
```diff
@@ -3,7 +3,12 @@ all:
       admin_groups: None
       admin_users: None
       computer_name: None
-      environment: False
+      environment:
+      valid_environments:
```
lorengordon: We can remove this from the default config. These values are implementation-specific, and we try not to impose implementation-specific values on users via the default config...

src/watchmaker/__init__.py
confusdcodr: This has been flattened and a before_install method has been added both to the individual workers and to the worker manager

src/watchmaker/static/config.yaml
lorengordon: I'm still seeing this change in the PR... This file should not be changed in this PR.

src/watchmaker/managers/base.py
```diff
@@ -375,10 +377,27 @@ def _worker_execution(self):
     def _worker_validation(self):
         pass
 
-    @abc.abstractmethod
-    def worker_cadence(self):  # noqa: D102
+    def _before_install(self, workers):
```
lorengordon: With the current approach, this `_before_install` method isn't doing anything? Seems reasonable to remove it?

src/watchmaker/workers/salt.py
```diff
@@ -89,7 +89,8 @@ def __init__(self, *args, **kwargs):
         # Pop arguments used by SaltBase
         self.user_formulas = kwargs.pop('user_formulas', None) or {}
         self.computer_name = kwargs.pop('computer_name', None) or ''
-        self.ent_env = kwargs.pop('environment', None) or ''
+        self.ent_env = kwargs.pop('environment', None) or None
```
lorengordon: What's the reasoning behind changing the default here from `''` to `None`?

src/watchmaker/workers/salt.py
```diff
@@ -89,7 +89,8 @@ def __init__(self, *args, **kwargs):
...
+        self.valid_envs = kwargs.pop('valid_environments', [None]) or [None]
```
lorengordon: The default for `pop` can just be `None`. Then the `or` will takeover. That way the argument default is set really in just the one place rather than two.

Though, it seems to me the default should be an empty list `[]`, not a list containing one item `[None]`.

```suggestion
        self.valid_envs = kwargs.pop('valid_environments', None) or []
```

src/watchmaker/workers/salt.py
```diff
@@ -111,6 +112,22 @@ def __init__(self, *args, **kwargs):
...
 
+    def before_install(self):
+        """Validate configuration before starting install."""
+        # Convert environment to lowercase
+        env = str(self.ent_env).lower()
+
+        # Convert all valid environment to lowercase
+        valid_envs = [str(x).lower() for x in self.valid_envs]
+
+        if env not in valid_envs:
```
lorengordon: For compatibility with an empty list default...

```suggestion
        if valid_envs and env not in valid_envs:
```

tests/test_watchmaker.py
```diff
@@ -3,15 +3,55 @@
...
+
+
+@pytest.fixture
+def default_config():
+    """Return default configuration for watchmaker."""
+    config_path = os.path.join(static.__path__[0], 'config.yaml')
+    config = None
+
+    with open(config_path, 'r') as stream:
+        try:
```
lorengordon: The try/except here isn't really doing anything...

```suggestion
def default_config():
    config_path = os.path.join(static.__path__[0], 'config.yaml')

    with open(config_path, 'r') as stream:
        return yaml.safe_load(stream)
```

tests/test_saltworker.py
```diff
@@ -0,0 +1,41 @@
...
+@pytest.fixture
+def saltworker_client():
+    """Return salt worker arguments."""
+    system_params = {}
+    salt_config = {}
+
+    return SaltBase(system_params, **salt_config)
+
+
+def test_valid_environments(saltworker_client):
```
lorengordon: I'd prefer to see the 4 test cases here split into separate test functions. The pytest output is much cleaner.

```suggestion
def test_valid_environments_default_values(saltworker_client):
    assertIsNone(saltworker_client.before_install())

def test_valid_environments_env_bogus(saltworker_client):
    saltworker_client.ent_env = 'bogus'
    with pytest.raises(InvalidValue):
        saltworker_client.before_install()

def test_valid_environments_env_missing(saltworker_client):
    saltworker_client.valid_envs = ['dev', 'test', 'prod']
    with pytest.raises(InvalidValue):
        saltworker_client.before_install()

def test_valid_environments_env_present(saltworker_client):
    saltworker_client.ent_env = 'dev'
    saltworker_client.valid_envs = ['dev', 'test', 'prod']
    assertIsNone(saltworker_client.before_install())
```

src/watchmaker/workers/salt.py
confusdcodr: Fixed

src/watchmaker/workers/salt.py
confusdcodr: Fixed

tests/test_watchmaker.py
confusdcodr: Fixed

tests/test_saltworker.py
confusdcodr: Test 3 is no longer valid with the check for an empty array. 3 tests have been written.

src/watchmaker/workers/salt.py
confusdcodr: Fixed

src/watchmaker/static/config.yaml
confusdcodr: Fixed.

docs/configuration.md
```diff
@@ -151,7 +154,7 @@ all:
       admin_groups: None
       admin_users: None
       computer_name: None
-      environment: False
+      environment:
```
lorengordon: Let's be explicit like the others and make this `None`

src/watchmaker/static/config.yaml
```diff
@@ -3,7 +3,7 @@ all:
       admin_groups: None
       admin_users: None
       computer_name: None
-      environment: False
+      environment:
```
lorengordon: Let's be explicit like the others and make this `None`

lorengordon:
go codebuild go

PR #745, opened 2019-01-09
Refactor to include WorkerBase

confusdcodr:
There was no WorkerBase class for Workers to work from. The managers had confusing names so it was changed to make it more clear as to what service it is providing.


src/watchmaker/managers/worker.py
```diff
@@ -0,0 +1,30 @@
+# -*- coding: utf-8 -*-
```
lorengordon: How about putting this in `workers/base.py`?

confusdcodr: I actually renamed base.py to be platform.py and separate out the three services (platform management, worker management, and worker) into their own files.

lorengordon: Understood, but this is not a manager. It is a new base class, for workers. To me, it seems more natural to put it in the `workers.base` module.

confusdcodr: Ah, misunderstood what you were saying. That's my bad. In my mind I was thinking managers/base.py

confusdcodr: Done

lorengordon:
go codebuild go

confusdcodr:
go codebuild go

PR #758, opened 2019-01-29
add amazon-inspector-formula as a submodule

confusdcodr:


confusdcodr:
Closing this pull request as I actually need to update the salt-formula and then readd the formula to the modules so that it is pointing to the correct commit

PR #790, opened 2019-02-13
Move deploy to Xenial

YakDriver:


YakDriver:
This allows Linux standalones to build properly again.

PR #810, opened 2019-02-28
Add sshd_config notes for 'MACs' and 'Ciphers' findings

### LINKED ISSUE #797: Add Content to Remove CBC Ciphers from SSH Daemon's Configuration
Some scanners' implementation of the `sshd_use_approved_ciphers` rule will complain that the `Ciphers` value is not an exact-match to:

    aes128-ctr,aes192-ctr,aes256-ctr

The `oscap`-based remediation used by watchmaker's [ash-linux-formula](https://github.com/plus3it/ash-linux-formula) components will typically install a string of:

    aes128-ctr,aes192-ctr,aes256-ctr,aes128-cbc,3des-cbc,aes192-cbc,aes256-cbc

As these are all Ciphers that the vendor was able to certify under FIPS-140-2. To prevent false alerts, likely need to trim the Ciphers-list.

Note: this wasn't one of the originally-noted findings in #704, but was uncovered as part of the process of addressing _those_.

### LINKED ISSUE #800: Add Content to Remove "Extra" `MACs` From SSH Daemon Configuration
Some scanners' implementation of the `sshd_use_approved_macs` rule will complain that the `MACs` value is not an exact-match to:

    hmac-sha2-512,hmac-sha2-256

The `oscap`-based remediation used by watchmaker's [ash-linux-formula](https://github.com/plus3it/ash-linux-formula) components will typically install a string of:

    hmac-sha2-512,hmac-sha2-256,hmac-sha1,hmac-sha1-etm@openssh.com,hmac-sha2-256-etm@openssh.com,hmac-sha2-512-etm@openssh.com

As these are all MACs that the vendor was able to certify under FIPS-140-2. To prevent false alerts, likely need to trim the MAC-list.

Note: this wasn't one of the originally-noted findings in #704, but was uncovered as part of the process of addressing _those_.

ferricoxide:
* Watchmaker sets correct values, however, vendor-validation is incorrectly marking the values as non-compliant
* Added a section for sshd_config settings since more than one setting is needs to be explained

Validated that the technical issues 266 and 269 were addressed in ash-linux-formula project. However, since the utility is still spuriously declaring an error, wanted to document same, here (and link the document-sections out to the BZ). Should close #797 and close #800 (in this project).

### DIFF
--- a/docs/findings/el7.md
+++ b/docs/findings/el7.md
@@ -7,7 +7,9 @@
 .. |sshdconfig| replace:: ``sshd_config (5)``
 .. _sshdconfig: https://linux.die.net/man/5/sshd_config
 
-## `LogLevel` Parameter's value not set to `INFO` in `/etc/ssh/sshd_config`
+## `sshd_config` Parameter Values
+
+### `LogLevel` Parameter's Value Not Set to `INFO` in `/etc/ssh/sshd_config`
 
 The "stock" `/etc/ssh/sshd_config` file typically contains a commented-out line for the `LogLevel` parameter similar to the following:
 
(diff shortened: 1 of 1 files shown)


lorengordon:
@ferricoxide you want to hold off on this one, while @eemperor tests the newer upstream scap content? that content may include the patch that addresses the issue.

ferricoxide:
> @ferricoxide you want to hold off on this one, while @eemperor tests the newer upstream scap content? that content may include the patch that addresses the issue.

I looked at the upstream project before opening the new BZ. The regexes are the same between upstream and vendor.

lorengordon:
oh i see now, the patch that presumably fixed/closed the issue was released in v0.1.32. if it's still not working... then yeah, we'll just keep getting false positives.

ferricoxide:
Yeah. It's annoying. And, looking at the regex, it looks like it _should_ work. I even tried changing the ordering of the scanned for parms in case the regex was expecting a specific order, but, "no dice". :-\

PR #814, opened 2019-03-04
Allow for integrated salt content

YakDriver:


YakDriver:
* Move salt-content from https://github.com/plus3it/watchmaker-content to watchmaker repo: `src/watchmaker/static/salt/content`.
* Change config so that when `salt_content: None`, Watchmaker uses base repo content (`src/watchmaker/static/salt/content`)
* Functions as before if `salt_content` attribute is something other than `None`.

YakDriver:
The default `config.yaml` is this PR has `salt_content: None`. To test this with CodeBuild, I change the `TF_VAR_tfi_common_args="-n -e dev"`, removing the `-c` pointing to a `config.yaml` that points to a Salt content zip file.

After this is worked out, we should be able to archive the watchmaker-content repo.

YakDriver:
go codebuild go

src/watchmaker/workers/salt.py
```diff
@@ -5,6 +5,7 @@
 
 import ast
 import codecs
+import distutils
```
lorengordon: Can we use `shutil.copytree`, rather than introducing a new import?

src/watchmaker/workers/salt.py
```diff
@@ -241,7 +242,7 @@ def _get_formulas_conf(self):
         ]
 
     def _build_salt_formula(self, extract_dir):
-        if self.salt_content:
+        if self.salt_content and self.salt_content != 'None':
```
lorengordon: What is the extra check here protecting us from?

src/watchmaker/workers/salt.py
```diff
@@ -254,6 +255,12 @@ def _build_salt_formula(self, extract_dir):
                 filepath=salt_content_file,
                 to_directory=extract_dir
             )
+        else:
+            local_content = os.sep.join((
+                static.__path__[0], 'salt', 'content'))
+            if os.path.exists(local_content):
+                self.log.info('Using local salt content: %s', local_content)
```
lorengordon: This wording is slightly confusing, since we're "using" it only in the sense that it is being copied to the location salt expects (or will expect) in the salt config. _Salt_ is not "using" this location. When users run watchmaker, they're generally unaware of this kind of distinction, and I think they would assume this log entry would mean the latter.

src/watchmaker/workers/salt.py
```diff
@@ -254,6 +255,12 @@ def _build_salt_formula(self, extract_dir):
...
+                distutils.dir_util.copy_tree(local_content, extract_dir)  # noqa: E501 # pylint: disable=no-member,line-too-long
```
lorengordon: Why is pylint complaining about `no-member` here?

src/watchmaker/workers/salt.py
YakDriver: `shutil.copytree` is picky: 

> The destination directory, named by _dst_, must not already exist; it will be created.

Here `extract_dir` exists when `shutil.copytree` would be used. `dir_util.copy_tree` does not require that the destination directory not exist. If _dst_/`extract_dir` is not important, I can remove it and recreate with `shutil.copytree`.

src/watchmaker/workers/salt.py
YakDriver: If `salt_content` is set to `None` in `config.yaml`, execution arrives at `_build_salt_formula()` with `self.salt_content` equal to the string `None`. It then tries to copy and extract a zip called `None`. This limited code changes elsewhere.

src/watchmaker/workers/salt.py
YakDriver: 👍

src/watchmaker/workers/salt.py
YakDriver: I wish I knew! https://stackoverflow.com/questions/18908941/unable-to-import-distutils-dir-util-on-windows/33417555

src/watchmaker/workers/salt.py
lorengordon: I think we _must_ remove the directory. That we're not may be a latent bug. Would need to trace through the setup of the salt srv directory structure to be sure.

src/watchmaker/workers/salt.py
lorengordon: Another reason not to use `distutils` 😛

src/watchmaker/workers/salt.py
lorengordon: Ahh, right, we need to fix that too. Separately, though. This kind of special handling for `'None'` is basically a user-error (ours). In yaml, it should be either an empty assignment or `null`... `salt_content:` or `salt_content: null`.

src/watchmaker/workers/salt.py
YakDriver: Will do. 👍

PR #816, opened 2019-03-05
WIP: Allow for local salt content

YakDriver:


YakDriver:
go codebuild go

YakDriver:
* Move salt-content from https://github.com/plus3it/watchmaker-content to watchmaker repo: `src/watchmaker/static/salt/content`.
* Change config so that when `salt_content: None`, Watchmaker uses included base repo content (src/watchmaker/static/salt/content)
* Works as before if `salt_content` attribute is something other than `None`.

The default `config.yaml` in this PR has `salt_content: None`. To test using Watchmaker-included content with CodeBuild, change the TF_VAR_tfi_common_args="-n -e dev", removing the -c pointing to a external config.yaml that points to an external Salt content zip file.

After this is worked out, we should be able to archive https://github.com/plus3it/watchmaker-content.

src/watchmaker/workers/salt.py
```diff
@@ -240,8 +240,17 @@ def _get_formulas_conf(self):
                 self.salt_formula_root))[1]
         ]
 
+    def _clear_copy_dir(self, src, dst, sub):
+        source = os.sep.join((src, sub))
+        destination = os.sep.join((dst, sub))
+        if os.path.exists(source):
```
lorengordon: I think if `source` doesn't exist, that's a valid error condition. We don't need to check for it, just let it raise the exception. Otherwise we get into a weird situation where it's possible for the method to "complete" without actually copying anything, and without letting the user handle the condition.

YakDriver: 👍

src/watchmaker/workers/salt.py
```diff
@@ -240,8 +240,16 @@ def _get_formulas_conf(self):
                 self.salt_formula_root))[1]
         ]
 
+    def _clear_copy_dir(self, src, dst, sub):
```
lorengordon: I think it makes more sense to handle the `sub` outside this method, as we don't really need `src` or `dst` otherwise. Function signature would be more clear as just `def _clear_copy_dir(src, dst):` (since `self` isn't used either)

confusdcodr: @YakDriver, I don't see any tests :|

YakDriver: @confusdcodr Challenge accepted.

src/watchmaker/workers/salt.py
```diff
@@ -20,6 +20,12 @@
 from watchmaker.workers.base import WorkerBase
 
 
+def _clear_copy_dir(src, dst):
```
lorengordon: If we're going to move this out of the class, then please put it in the `utils` module.

Also, what do you think about overloading the shutil `copytree` function? i.e....

```
def copytree(src, dst, force=False, **kwargs):
    if os.path.exists(dst) and force:
        shutil.rmtree(dst)
    shutil.copytree(src, dst, **kwargs)
```

YakDriver: I moved this to utils which made testing much easier. All I changed was switching the order of `if` conditions. I did not override the shutil function but used the same signature. Is that what you wanted?

YakDriver:
This is an iterative step in the right testing direction IMHO but definitely not a definitive, comprehensive suite. What tests this includes are mostly simple rather than fully probing all logical branches and problem spots.

lorengordon:
go codebuild go

setup.cfg
```diff
@@ -39,7 +39,7 @@ install_requires =
     six
     pypiwin32;platform_system=="Windows"
     PyYAML;python_version>="2.7"
-    PyYAML<4;python_version<"2.7"    
+	PyYAML<4;python_version<"2.7"
```
lorengordon: Missing a space here...

YakDriver: silly vscode 😞 - using tabs instead of spaces

YakDriver:
go codebuild go

YakDriver:
Regular codebuild isn't too useful for testing this PR since its config doesn't have `salt_content: None`. I ran on my system with args of `-n -e dev` to use included config which does have `salt_content: None`. It succeeded.

YakDriver:
go codebuild go

src/watchmaker/utils/__init__.py
```diff
@@ -58,3 +59,37 @@ def urlopen_retry(uri):
...
+            (*Default*: ``False``)
+
+        Returns:
+            :obj:`bool`:
+                ``True`` if the copy was attempted. ``False`` if not.
+
+    """
+    if force and os.path.exists(dst):
+        shutil.rmtree(dst)
+    if not ignore or not os.path.exists(dst):
```
lorengordon: This `ignore` seems a bit tortured. We're calling a function whose purpose is to copy a directory, but telling the function _not_ to copy the directory? This is the wrong place for that logic.

src/watchmaker/workers/salt.py
```diff
@@ -255,6 +260,22 @@ def _build_salt_formula(self, extract_dir):
                 to_directory=extract_dir
             )
 
+        bundled_content = os.sep.join(
+            (static.__path__[0], 'salt', 'content')
+        )
+        if os.path.exists(bundled_content):
```
lorengordon: This can't _not_ exist. That's an exception, let it raise.

src/watchmaker/workers/salt.py
```diff
@@ -255,6 +260,22 @@ def _build_salt_formula(self, extract_dir):
...
+            for subdir in next(os.walk(bundled_content))[1]:
+                if not subdir.startswith('.') and watchmaker.utils.copytree(
```
lorengordon: Yeah, rewrite this so the `if` is just a test. Setup the test, and if the test passes then run `copytree()`.

src/watchmaker/workers/salt.py
```diff
@@ -241,7 +240,13 @@ def _get_formulas_conf(self):
         ]
 
     def _build_salt_formula(self, extract_dir):
-        if self.salt_content:
+        force_copy_bundled_content = True
```
lorengordon: Figure we can get rid of these two flags as well.

lorengordon:
Superseded by #880

PR #863, opened 2019-04-19
Restricts backoff version to retain compatibility with python2.6

lorengordon:


lorengordon:
Here's the changelog indicating the backoff version that dropped py2.6 support...

* https://github.com/litl/backoff/blob/master/CHANGELOG.md#v170---2018-11-23

PR #884, opened 2019-05-09
Updates gitlab-ci to use python36 and ensurepip

lorengordon:


.gitlab-ci.yml
```diff
@@ -1,8 +1,9 @@
 image: ${IMAGE}
 
 before_script:
-  - sudo yum -y install python34 python34-pip python34-libs python34-setuptools
-  - sudo python3 -m pip install --upgrade pip
+  - sudo yum -y install python36 python36-libs python36-setuptools
+  - sudo python3.6 -m ensurepip --upgrade
```
eemperor: Shouldn't this be `python3`?

lorengordon: No, the `python3*` packages are no longer creating the `python3` symlink, just the explictly-versioned executables... We could add another line that created the symlink if we wanted, but gitlab-ci runners are generally ephemeral so there isn't much point in making the environment too user-friendly.

eemperor: I see.  Good with me.

PR #885, opened 2019-05-09
Adds Python 3.7 support to CI

eemperor:


eemperor:
Oops.  Guess I should do what comment says...

.travis.yml
```diff
@@ -54,6 +54,16 @@ notifications:
     on_failure: always
 jobs:
   include:
+    - stage: test
+      python: 3.7
+      dist: xenial
+      env:
+        - TOXENV=check
```
lorengordon: I think we can remove the `env` block. We don't want to override TOXENV. It should be set by the top-level `before_install`.

.travis.yml
```diff
@@ -54,6 +54,16 @@ notifications:
...
+    - stage: test
```
lorengordon: This job can be removed. We aren't worried at the moment about the docs. We're really just getting at pytest.

.travis.yml
```diff
@@ -87,7 +97,7 @@ jobs:
           tags: true
     - stage: deploy
       if: branch in (env(RELEASE_BRANCH), develop) OR tag is present
-      python: 3.6
+      python: 3.7
```
lorengordon: Let's not change this... I think that will impact the standalone builds.

appveyor.yml
```diff
@@ -14,10 +14,10 @@ init:
 install:
   - git submodule update --init --recursive
 test_script:
-  # ----- python 3.6, x64
+  # ----- python 3.7, x64
```
lorengordon: This also should be a _new_ test, not replacing the 3.6 tests...

eemperor:
@lorengordon Updates made per your comments.  Ready for re-review.

PR #928, opened 2019-07-24
Add Azure Pipelines CI

YakDriver:


azure-pipelines.yml
```diff
@@ -0,0 +1,152 @@
...
+        - task: S3Upload@1
+          displayName: upload standalone binaries to s3 (release)
+          condition: eq(variables['build.sourceBranch'], 'refs/tags/*')
+          inputs:
+            awsCredentials: dicelab
+            targetFolder: $(keyPrefix)
+            #regionName: 'us-east-1'
+            bucketName: $(releaseBucket)
+            sourceFolder: $(deployStage)
+        - script: |
```
lorengordon: Do we know that non-zero return codes from individual commands in these multi-line script-blocks will cause the script-block to fail? Or, like in bash or powershell by default, will it continue on to the next command, potentially causing a false success or otherwise obscuring the original error?

PR #951, opened 2019-08-21
Adds submodule for fup-formula

ferricoxide:


.gitmodules
```diff
@@ -46,3 +46,6 @@
 [submodule "src/watchmaker/static/salt/content"]
 	path = src/watchmaker/static/salt/content
 	url = https://github.com/plus3it/watchmaker-salt-content.git
+[submodule "src/watchmaker/static/salt/formulas/fup-formula"]
+	path = src/watchmaker/static/salt/formulas/fup-formula
+	url = git@github.com:plus3it/fup-formula.git
```
lorengordon: eh, let's use https remotes, like all the rest of the submodules...

ferricoxide: Oop. Sorry. Habit (and wasn't specified in the doc 😛)

Fixing, shortly.

PR #980, opened 2019-09-24
Fix condition so release jobs trigger on tags

YakDriver:


YakDriver:
Next tag, it should work! 🤞 (It does in my test repo anyway.)

PR #987, opened 2019-10-04
Add version to config, deprecate None, remove wheel use of m2r

YakDriver:


YakDriver:
go codebuild go

YakDriver:
codebuild is failing because RHEL6 from source is failing because "Service postfix failed to start."

YakDriver:
go codebuild go

YakDriver:
go codebuild go

YakDriver:
go codebuild go

src/watchmaker/__init__.py
```diff
@@ -301,6 +304,12 @@ def _get_config(self):
             self.log.critical(msg)
             raise WatchmakerException(msg)
 
+        if config_version_specifier and not check_version(
+                watchmaker.__version__, config_version_specifier):
+            msg = 'Watchmaker is not compatible with the config file'
```
lorengordon: I think it would be helpful to output the version and the constraint in the msg...
```suggestion
            msg = 'Watchmaker version {} is not compatible with the config file (watchmaker_version = {})'.format(watchmaker.__version__, config_version_specifier)
```

src/watchmaker/static/config.yaml
```diff
@@ -1,11 +1,12 @@
+watchmaker_version: "== 0.16.6"
```
lorengordon: How are we maintaining this pin? I think this can at least be:
```suggestion
watchmaker_version: ">= 0.17"
```

src/watchmaker/__init__.py
```diff
@@ -301,6 +304,12 @@ def _get_config(self):
...
+            self.log.critical(msg)
+            raise WatchmakerException(msg)
+
         # Merge the config data, preserving the listed order of workers
         config = collections.OrderedDict()
         for worker in config_system + config_all:
```
lorengordon: I'm not sure if it's worth it, but somewhere in here we _could_ check for string `'None'` values, force them to `None`, and log/raise a [`DeprecationWarning`](https://docs.python.org/3/library/warnings.html#warning-categories)...

src/watchmaker/static/config.yaml
YakDriver: I used `~= 0.17.0` which will limit to 0.17.x.

src/watchmaker/__init__.py
YakDriver: This should not be too hard. I'll get it in.

YakDriver: Take a look at my approach to this and see what you think.

YakDriver: My approach is to centralize the handling of `None`. Since it is centralized, it will be easy to switch to raising an error and then phasing out.

YakDriver:
go codebuild go

src/watchmaker/utils/__init__.py
```diff
@@ -83,3 +84,31 @@ def copytree(src, dst, force=False, **kwargs):
...
+    r"""
+    Check whether a variable is the string 'None' rather than Pythonic `None`.
+
+    If it is the string 'None', this warns and returns Pythonic `None`.
+    Otherwise, the variable is returned unchanged.
+
+    Args:
+        check_value: (:obj:`str`)
+            Variable to be checked for string 'None' (case-insenstive).
+            (*Default*: None)
```
lorengordon: I think there is no default value, both these params are required...

src/watchmaker/utils/__init__.py
```diff
@@ -83,3 +84,31 @@ def copytree(src, dst, force=False, **kwargs):
...
+            Logger where deprecation warning will be made.
+            (*Default*: None)
+
+    """
+    if check_value and check_value.lower() == 'none':
+        deprecate_msg = (
+            'Use of "None" (string) as a config value is deprecated. Use '
+            '`null` instead.')
+        log.warn(deprecate_msg)
+        warnings.warn(deprecate_msg, DeprecationWarning)
```
lorengordon: Do we get two log outputs when we do both log.warn and warnings.warn?

src/watchmaker/workers/salt.py
```diff
@@ -151,10 +151,10 @@ def _prepare_for_install(self):
             self.salt_working_dir_prefix
         )
 
-        if (
-            self.salt_debug_log and
-            self.salt_debug_log != 'None'
-        ):
+        self.salt_debug_log = watchmaker.utils.config_none_deprecate(
```
lorengordon: How about moving these to the `__init__` where we set `self.salt_debug_log` in the first place?

src/watchmaker/workers/salt.py
```diff
@@ -477,14 +489,16 @@ def process_states(self, states, exclude):
                 Comma-separated string of salt states to execute. Accepts two
                 special keywords (case-insensitive):
 
-                - ``none``: Do not apply any salt states.
+                - ``null``: Do not apply any salt states.
```
lorengordon: This one may need special handling. I think this impacts the cli usage of `--salt-states None`. Does that still work, and just outputs the log warning?

src/watchmaker/__init__.py
lorengordon: Good enough for now, until we get a real config spec/parser

src/watchmaker/utils/__init__.py
YakDriver: No. `warnings.warn`'s behavior can be customized to do things like raising an error, filter out warnings, or possibly log the error. The default is to output to stderr. Figured this would be another way to gradually ramp up the deprecation consequences.

src/watchmaker/utils/__init__.py
YakDriver: Yes, absolutely. I misread. I was not thinking Pythonic `None` but rather that there was no default.

src/watchmaker/workers/salt.py
YakDriver: Absolutely. Makes more sense.

src/watchmaker/workers/salt.py
YakDriver: `null` is standard YAML but for the CLI, I think that `"None"` or `"none"` should be valid and not deprecated. But, we won't know whether it comes from YAML or the CLI.

lorengordon:
go codebuild go

YakDriver:
go codebuild go

src/watchmaker/workers/salt.py
```diff
@@ -452,16 +465,16 @@ def process_grains(self):
         self._set_grain('watchmaker', ent_env)
 
         grain = {}
-        if self.ou_path and self.ou_path != 'None':
+        if self.ou_path:
             grain['oupath'] = self.ou_path
-        if self.admin_groups and self.admin_groups != 'None':
+        if self.admin_groups and hasattr(self.admin_groups, "split"):
```
lorengordon: i don't think we need the suspenders here. `self.admin_groups` should be a string by this point or `None`

src/watchmaker/workers/salt.py
```diff
@@ -452,16 +465,16 @@ def process_grains(self):
...
             grain['admin_groups'] = self.admin_groups.split(':')
-        if self.admin_users and self.admin_users != 'None':
+        if self.admin_users and hasattr(self.admin_users, "split"):
```
lorengordon: same as `self.admin_groups`

src/watchmaker/workers/salt.py
```diff
@@ -568,7 +581,8 @@ def __init__(self, *args, **kwargs):
         self.salt_version = kwargs.pop('salt_version', None) or ''
 
         # Enforce lowercase and replace spaces with ^ in Linux
-        self.admin_groups = self.admin_groups.lower().replace(' ', '^')
+        if hasattr(self.admin_groups, 'lower'):
```
lorengordon: ok, `None` would be a problem. coerce the default to the expected type in the `__init__` method... e.g.

```
        self.admin_groups = watchmaker.utils.config_none_deprecate(
            self.admin_groups, self.log) or ''
```

lorengordon:
how about adding unit tests around the changes introduced here?

src/watchmaker/workers/salt.py
YakDriver: I like Pythonic `None` better than an empty string. This should be fixable with just checking to see if `admin_groups`. If so, it will be a string.

src/watchmaker/workers/salt.py
YakDriver: Done

src/watchmaker/workers/salt.py
YakDriver: Done

YakDriver:
@lorengordon I've added tests. This is ready for another look.

PR #1035, opened 2019-12-04
Prepares watchmaker 0.16.6 release

lorengordon:


lorengordon:
go codebuild go

moskey71:
Should be good so long as the codebuild check passes.

PR #1055, opened 2020-01-06
Restricts pyyaml version to support python 3.4 (for el6)

lorengordon:
This will be a patch release, to fix installs on EL6 using python 3.4...


lorengordon:
go codebuild go

PR #1059, opened 2020-01-10
Uses salt 2018.3.4 in the default config

lorengordon:
Also, this is the first version that properly detects Windows Server 2019


lorengordon:
go codebuild go

PR #1079, opened 2020-01-22
Merge pull request #1079 from lorengordon/none-salt

lorengordon:
This case is now handled in the `Arguments()` class for any
string 'None' on the cli.


lorengordon:
go codebuild go

PR #1090, opened 2020-02-05
Improves TOC handling by switching from m2r to recommonmark

lorengordon:


lorengordon:
I enabled RTD on my fork so I could build the docs and provide a preview... https://watchmaker-lorengordon.readthedocs.io/en/index-ext/index.html

YakDriver:
On installation, the first section got muffed. Other than that, looks good.

lorengordon:
Good catch! The `.. note::` RST block needed to be encapsulated in `eval_rst` blocks... I just pushed an update that I think accounts for all the prior special handling of RST we had in there that was relying on `m2r` functionality.

PR #1094, opened 2020-02-06
Documents configuration vs cli argument handling and precedence

lorengordon:


lorengordon:
Rendered:

* https://watchmaker-lorengordon.readthedocs.io/en/doc/configuration.html
* https://watchmaker-lorengordon.readthedocs.io/en/doc/usage.html

PR #1095, opened 2020-02-06
Known findings

ferricoxide:
Update "Known Findings" content to better align with current scan/remediate content and to include an easy-reference table with anchor-links to explanatory text.


docs/findings/el7.md
```diff
@@ -7,199 +7,89 @@
...
-grep -E '\s\/tmp\s.*nosuid' /proc/mounts
-grep nosuid /etc/systemd/system/tmp.mount.d/options.conf
-```
+Conditionally-Valid Finding: Smart Card Login use and configuration is site-specific. Site has not provided specification for implementing this setting within scanned context.
 
-The above verifies that the mounted `/tmp` has the desired mount-option set and that the setting will persist after a reboot.
+## Configure the Firewalld Ports
 
-### `/var/tmp` — `nodev` Option Missing
+Invalid Finding: Watchmaker implements setting. Howeve, scanner regex may not be sufficiently-flexible in its specification.
```
lorengordon:
```suggestion
Invalid Finding: Watchmaker implements setting. However, scanner regex may not be sufficiently-flexible in its specification.
```

docs/findings/el7.md
```diff
@@ -7,199 +7,89 @@
...
 
-Cannot satisfy: `/var/tmp` not implemented as stand-alone filesystem.
+## Set Default firewalld Zone for Incoming Packets
```
lorengordon: I think we can remove this. We'd consider this more a bug/feature to fix, rather than something users should expect as a finding.

docs/findings/el7.md
```diff
@@ -7,199 +7,89 @@
...
+## Enable FIPS Mode in GRUB2
 
-Note: for programs that place the additional requirement that `/var/tmp` _must_ be a standalone filesystem, it is recommended to up the size of the root volume-group, create an LVM volume to host the `/var/tmp` filesystem and then apply the site-local mount-options to the filesystem. This can all be accomplished at system provisioning-time, but is outside the scope of the watchmaker utility to effect.
+Conditionally-Valid Finding: Both spel and watchmaker implement `fips=1` by default. If finding occurs, either:
 
-## Missing Audit Rules
+* There is an error in scanner's validation-method
+* System has been intentionally de-configured for FIPS &mdash; typically due to hosted-software's requirements &mdash; and will need to be POA&Med.
```
lorengordon:
```suggestion
* System has been intentionally de-configured for FIPS &mdash; typically due to hosted-software's requirements &mdash.
```

The POA&M term is very domain-specific. I think we should remove it or generalize.

docs/findings/el7.md
```diff
@@ -7,199 +7,89 @@
...
 
-Some third-paty scanners will incorrectly indicate that the `auditd` subsystem is missing rules.
+## Configure AIDE to Use FIPS 140-2 for Validating Hashes
```
lorengordon: Remove and address as a bug, rather than a known/expected finding.

docs/findings/el7.md
```diff
@@ -7,199 +7,89 @@
...
+## Verify and Correct Ownership with RPM
 
-### Ensure Attempts to Alter Logon and Logout Events Is Collected
+Invalid Finding:
 
-Already implemented in the `/etc/audit/audit.rules` and `/etc/audit/rules.d/logins.rules` files. Watchmaker-applied rule uses `-k` to signify the audit event-key. Complaining scanner-utility is likely searching for the `key=` token to signify the audit event-key or similarly-incompatible search-string.
+* Flags on system-journal ownership: Journal ownership settings are automatically reset by systemd (upon reboot) after hardening has run. Currently, no means of permanently remediating is possible.
+* Similarly, if HBSS or VSEL is installed, scan may flag on user-ownership depending on how site specifies installation of HBSS or VSEL. One would reasonably expect similar for other, third-party packages.  "Fixing" (per STIG guidance) would likely break the functioning of the HBSS/VSEL (or third-party) software
```
lorengordon: I think we should *focus* on the general cause here (installing packages using `rpm` instead of `yum`), and mention the McAfee Agent as a specific example, rather than the other way around.

docs/findings/el7.md
```diff
@@ -7,199 +7,89 @@
 ```
 <br>
 
-## `sshd_config` Parameter Values
+| Finding Summary | Finding Identifiers |
```
lorengordon: Let's add a header for the table also, maybe `# Finding Summary Table`?

lorengordon:
I'd also like to keep more of the original findings/el7.md content, as it was certainly covering valid topics. We just need to dig some to get finding IDs. Or maybe we just leave that column blank if we don't have a finding ID for something? Wouldn't be the end of the world...

docs/findings/el7.md
ferricoxide: "Short-term deployment security-exception"?

docs/findings/el7.md
ferricoxide: Using `rpm` vice `yum` isn't the issue, per se. It's a combination of:

* The relevant RPMs not declaring certain files as "config" (or even "ghost") files
* The scanner is using a faulty change-detection method - ignoring files that actually _are_ managed as "config" (or "ghost") files

docs/findings/el7.md
ferricoxide: > I'd also like to keep more of the original findings/el7.md content, as it was certainly covering valid topics. We just need to dig some to get finding IDs. Or maybe we just leave that column blank if we don't have a finding ID for something? Wouldn't be the end of the world...

Given the output of the stig-based `--remediate` and `--scan`, those appeared to no longer be valid. So, I nuked them out in favor of what's currently-returned. Since we have no reports relevant to using other profiles for  `--remediate` and `--scan`, had no other profile-specific known findings to reference (and, if did, would probably create sections for each relevant, not-retired profile)

ferricoxide:
> I'd also like to keep more of the original findings/el7.md content, as it was certainly covering valid topics. We just need to dig some to get finding IDs. Or maybe we just leave that column blank if we don't have a finding ID for something? Wouldn't be the end of the world...

Given the output of the stig-based `--remediate` and `--scan`, those appeared to no longer be valid. So, I nuked them out in favor of what's currently-returned. Since we have no reports relevant to using other profiles for  `--remediate` and `--scan`, had no other profile-specific known findings to reference (and, if did, would probably create sections for each relevant, not-retired profile)

docs/findings/el7.md
ferricoxide: Once I fix the `ash-linux-formula` content, I'd probably say that this is a conditional: it's a near lock that there's going to be sites that don't want the default zone to be "drop" &mdash; especially if they rely on ping-sweeps for anything (security-related or otherwise).

docs/findings/el7.md
```diff
@@ -7,199 +7,152 @@
 ```
 <br>
 
-## `sshd_config` Parameter Values
+# Findings Summary-Table
```
lorengordon: The navigation works better if this is a level 2 header...
```suggestion
## Findings Summary-Table
```

docs/findings/el7.md
lorengordon: This looks resolved now I think?

docs/findings/el7.md
lorengordon: HBSS isn't actually a package, right? It's really the McAfee Agent and McAfee VSEL?

lorengordon:
ok, this pr ought to fix the table width: https://github.com/plus3it/watchmaker/pull/1100

PR #1100, opened 2020-02-12
Overrides RTD theme so content width is responsive to screen size

lorengordon:


lorengordon:
virtualenv pushed a major update a couple days ago that broke some test stuff. Pinned the tox deps as a result. Please re-review @plus3it/watchmaker

lorengordon:
@plus3it/watchmaker one more look please. had to map out the dep/python version deprecations to get the pins to actually work. tests are all green _now_!

PR #1114, opened 2020-02-24
Uses a custom default so `None` cli arguments are passed to workers

lorengordon:
Previously, `None` was filtered from the arguments passed to workers
as a way of differentiating between arguments that were passed on
the cli and those that simply had a `None` default value. This
worked when the worker had special handling for the string `'None'`
because that is how the cli would pass the string value along.

When the special handling for `None` was deprecated, the filtering
of `None` from cli args began removing the arg even if the user
specified the value.

In other words, `--salt-states none` would get converted properly
from `{'salt_states': 'none'}` to `{'salt_states': None}`, but then
the filter would remove the arg instead of passing it to the worker.

This patch changes the default for these worker args to a static
value, `Arguments.DEFAULT_VALUE`, and then filters that value
instead of `None`. This way users can pass `'None'`, the
Arguments() processing can change it to `None`, and it will still
get passed to the worker.

### DIFF
--- a/src/watchmaker/__init__.py
+++ b/src/watchmaker/__init__.py
@@ -167,6 +167,8 @@ class Arguments(dict):
 
     """
 
+    DEFAULT_VALUE = 'WAM_NONE'
+
     def __init__(
         self,
         config_path=None,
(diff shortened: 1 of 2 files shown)


lorengordon:
go codebuild go

lorengordon:
go codebuild go

lorengordon:
go codebuild go

PR #1121, opened 2020-02-26
Fix secret token passing to Satsuki

YakDriver:


azure-pipelines.yml
```diff
@@ -146,3 +146,5 @@ jobs:
             satsuki
           displayName: deploy to github with satsuki
           condition: startsWith(variables['build.sourceBranch'], 'refs/tags/')
+          env:
+            SATS_TOKEN: $(githubToken)
```
lorengordon: What happens if you `echo $SATS_TOKEN` in the `script`? Does it actually output the value? That would be ungood...

YakDriver: No. The docs specifically show an example of that. It just shows stars. https://docs.microsoft.com/en-us/azure/devops/pipelines/process/variables?view=azure-devops&tabs=yaml%2Cbatch#secret-variables

PR #1149, opened 2020-03-12
Point Satsuki to release asset information

YakDriver:


YakDriver:
@lorengordon This repo will automatically release a new version if there's a commit. I like letting it do it's thing but if you prefer a specific bump, I can do that too.

PR #1180, opened 2020-03-23
Updates version restriction in default config

### LINKED ISSUE #1179: Config.yaml not updated with v0.18.0 release
The static config.yaml file here:
* https://github.com/plus3it/watchmaker/blob/master/src/watchmaker/static/config.yaml

still references v0.17.0

lorengordon:
Fixes #1179

### DIFF
--- a/src/watchmaker/static/config.yaml
+++ b/src/watchmaker/static/config.yaml
@@ -1,4 +1,4 @@
-watchmaker_version: "~= 0.17.0"
+watchmaker_version: ">= 0.17.0"
 all:
   - salt:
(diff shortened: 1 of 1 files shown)


lorengordon:
The codebuild job uses its own custom config.yaml, which is why this was not caught earlier.

Exercising a manual run here, using the default config.yaml: https://console.aws.amazon.com/codesuite/codebuild/projects/terrafirm/build/terrafirm%3A8fad23b3-55ba-4eb0-913e-9918dd31956f/log?region=us-east-1

PR #1197, opened 2020-04-08
Add util to secure files

YakDriver:


lorengordon:
Did I see that `oschmod` is GPLv3?

> However, GPLv3 software cannot be included in Apache projects. The licenses are incompatible in one direction only, and it is a result of ASF's licensing philosophy and the GPLv3 authors' interpretation of copyright law.

https://www.apache.org/licenses/GPL-compatibility.html

YakDriver:
Bugger, I'll fix it.

YakDriver:
@lorengordon Fixed. It is now Apache 2.

setup.cfg
```diff
@@ -47,6 +47,7 @@ install_requires =
     PyYAML<4;python_version<"2.7"
     wheel<=0.29.0;python_version<"2.7"
     compatibleversion>=0.1.2
+    oschmod
```
lorengordon, line 50: To ensure we remain license-compatible...
```suggestion
    oschmod>=0.1.3
```

lorengordon:
@eemperor can you run from this branch and confirm it addresses the problem?

eemperor:
okay.  i'll test it.

src/watchmaker/utils/__init__.py
```diff
@@ -119,3 +121,34 @@ def clean_none(value):
         return None
 
     return value
+
+
+def set_file_perms(path, dir_mode=None, file_mode=None):
```
lorengordon, line 126: maybe name this to reflect it is not operating on files?
```suggestion
def set_path_perms(path, dir_mode=None, file_mode=None):
```

src/watchmaker/utils/__init__.py
```diff
@@ -119,3 +121,34 @@ def clean_none(value):
...
+
+    """
+    if not dir_mode:
+        dir_mode = stat.S_IEXEC | stat.S_IWRITE | stat.S_IREAD
+
+    if not file_mode:
+        file_mode = stat.S_IWRITE | stat.S_IREAD
+
+    for root, dirs, files in os.walk(path, topdown=False):
+        for one_file in [os.path.join(root, f) for f in files]:
```
lorengordon, line 148: it seems a bit excessive to perform the join here. this is a loop in a loop in a loop. move the join to the call to `oschmod.set_mode` and it's just a loop in a loop...
```suggestion
        for one_file in files:
            oschmod.set_mode(os.path.join(root, one_file), file_mode)
```

same for the loop over `dirs`...

src/watchmaker/__init__.py
```diff
@@ -452,6 +452,12 @@ def install(self):
             self.log.critical(msg)
             raise
 
+        watchmaker.utils.set_file_perms(self.system_params['prepdir'])
```
lorengordon, line 455: Did you happen to try setting the mode on `self.system_params['prepdir']` just after

```
            os.makedirs(self.system_params['workingdir'])
```
on line 433?

I'm curious if the files would all just inherit the correct perms then, so we wouldn't need to loop over all the files/dirs to set them individually...

YakDriver: It won't work. On nix, there's no inheritance but I'll set the umask before anything is created (ln 410). On doze, oschmod specifically disables inheritance to improve security/greatly simplify permissions.

tests/test_utils.py
```diff
@@ -51,3 +58,45 @@ def test_clean_none():
...
+    # check it out
+    assert oct(oschmod.get_mode(topdir))[-3:] == "700"
+    assert oct(
+        oschmod.get_mode(os.path.join(topdir, 'testdir2')))[-3:] == "700"
+    assert oct(oschmod.get_mode(testdir))[-3:] == "700"
+    assert oct(oschmod.get_mode(os.path.join(topdir, 'file1')))[-3:] == "600"
+    assert oct(oschmod.get_mode(os.path.join(testdir, 'file2')))[-3:] == "600"
+
+    # clean up
+    shutil.rmtree(topdir)
```
lorengordon, line 102: do we have the ability to use setup/teardown features of pytest here? seems like the directory/file creation will get a little weird when there are exceptions

YakDriver: Ideally, it would be mocked. Or, since this is really something we are handing off to oschmod, we could just leave the testing there. It has testing but this util could be moved there.

src/watchmaker/logger/__init__.py
```diff
@@ -141,6 +143,8 @@ def prepare_logging(log_dir, log_level):
         make_log_dir(log_dir)
         log_filename = os.sep.join((log_dir, 'watchmaker.log'))
         hdlr = logging.FileHandler(log_filename)
+        oschmod.set_mode(log_filename, 0o600)
+        oschmod.set_mode(log_dir, 0o700)
```
lorengordon, line 147: I don't think we should explicitly set the mode on `log_dir` here. This is a user-provided argument. The user should manage the permissions directly on whatever directory they supply. Say for example they pass `--log-dir /var/logs` -- it would be out of place for watchmaker to set the mode there.

Or alternatively, only set the permissions if watchmaker _creates_ the directory, in `make_log_dir`.

tests/test_utils.py
lorengordon: maybe that makes more sense. it's really like recursive mode for `chmod`, but with separate arguments for directories vs files. maybe extend `oschmod.set_mode`, or add a `oschmod.set_mode_recursive` function?

src/watchmaker/logger/__init__.py
YakDriver: Was just coming to the same conclusion

YakDriver:
Let me know if there's anything else. Otherwise, this should be ready to go.

lorengordon:
go codebuild go

PR #1205, opened 2020-04-15
Adding new arg 'salt_content_path' to specify salt files location in 'salt_content' archive file.

spavuluri:
Added new argument called `salt_content_path` to specify the location of Salt configuration files within the archive file specified by the `salt_content` arg.

### DIFF
--- a/src/watchmaker/utils/__init__.py
+++ b/src/watchmaker/utils/__init__.py
@@ -121,7 +121,7 @@ def clean_none(value):
     return value
 
 
-def copy_subdirectories(src_dir, dest_dir, log):
+def copy_subdirectories(src_dir, dest_dir, log=None):
     """Copy subdirectories within given src dir into dest dir."""
     for subdir in next(os.walk(src_dir))[1]:
         if (
--- a/src/watchmaker/utils/__init__.py
+++ b/src/watchmaker/utils/__init__.py
@@ -119,3 +119,20 @@ def clean_none(value):
         return None
 
     return value
+
+
+def copy_subdirectories(src_dir, dest_dir, log):
+    """Copy subdirectories within given src dir into dest dir."""
+    for subdir in next(os.walk(src_dir))[1]:
+        if (
+            not subdir.startswith('.') and
+            not os.path.exists(os.sep.join((dest_dir, subdir)))
+        ):
+            copytree(
+                os.sep.join((src_dir, subdir)),
+                os.sep.join((dest_dir, subdir))
+            )
+            if log:
+                log.info('Copied from %s to %s',
+                         os.sep.join((src_dir, subdir)),
+                         os.sep.join((dest_dir, subdir)))
--- a/src/watchmaker/workers/salt.py
+++ b/src/watchmaker/workers/salt.py
@@ -40,6 +40,13 @@ class SaltBase(WorkerBase, PlatformManagerBase):
             - *Linux*: ``/srv/watchmaker/salt``
             - *Windows*: ``C:\Watchmaker\Salt\srv``
 
+        salt_content_path: (:obj:`str`)
+            Used in conjunction with the "salt_content" arg.
+            Path to the salt content files inside the provided salt_content
+            archive. To be used when salt content files are located within a
+            sub-path of the archive, rather than at its top-level.
+            (*Default*: ``''``)
+
         salt_states: (:obj:`str`)
             Comma-separated string of salt states to execute. Accepts two
             special keywords (case-insensitive).
@@ -97,6 +104,7 @@ class SaltBase(WorkerBase, PlatformManagerBase):
         self.valid_envs = kwargs.pop('valid_environments', []) or []
         self.salt_debug_log = kwargs.pop('salt_debug_log', None) or ''
         self.salt_content = kwargs.pop('salt_content', None) or ''
+        self.salt_content_path = kwargs.pop('salt_content_path', None) or ''
         self.ou_path = kwargs.pop('ou_path', None) or ''
         self.admin_groups = kwargs.pop('admin_groups', None) or ''
         self.admin_users = kwargs.pop('admin_users', None) or ''
@@ -109,6 +117,8 @@ class SaltBase(WorkerBase, PlatformManagerBase):
             self.salt_debug_log, self.log)
         self.salt_content = watchmaker.utils.config_none_deprecate(
             self.salt_content, self.log)
+        self.salt_content_path = watchmaker.utils.config_none_deprecate(
+            self.salt_content_path, self.log)
         self.ou_path = watchmaker.utils.config_none_deprecate(
             self.ou_path, self.log)
         self.admin_groups = watchmaker.utils.config_none_deprecate(
@@ -266,6 +276,24 @@ class SaltBase(WorkerBase, PlatformManagerBase):
                 filepath=salt_content_file,
                 to_directory=extract_dir
             )
+            if self.salt_content_path:
+                self.log.debug(
+                    'Using salt content path: %s',
+                    self.salt_content_path
+                )
+                try:
+                    salt_content_src = os.sep.join((
+                        extract_dir, self.salt_content_path))
+                    salt_file_names = os.listdir(salt_content_src)
+                    for name in salt_file_names:
+                        srcname = os.sep.join((salt_content_src, name))
+                        shutil.move(srcname, extract_dir)
+                except OSError:
+                    if not os.path.isdir(salt_content_src):
+                        msg = 'Non-existent path in salt content ' \
+                              'file - {0}'.format(salt_content_src)
+                        self.log.critical(msg)
+                        raise
 
         bundled_content = os.sep.join(
             (static.__path__[0], 'salt', 'content')
--- a/src/watchmaker/workers/salt.py
+++ b/src/watchmaker/workers/salt.py
@@ -43,9 +43,12 @@ class SaltBase(WorkerBase, PlatformManagerBase):
 
         salt_content_path: (:obj:`str`)
             Used in conjunction with the "salt_content" arg.
-            Path to the salt content files inside the provided salt_content
-            archive. To be used when salt content files are located within a
-            sub-path of the archive, rather than at its top-level.
+            Glob pattern for the location of salt content files inside the
+            provided salt_content archive. To be used when salt content files
+            are located within a sub-path of the archive, rather than at its
+            top-level. Multiple paths matching the given pattern will result
+            in error.
+            E.g. ``salt_content_path='*/'``
             (*Default*: ``''``)
 
         salt_states: (:obj:`str`)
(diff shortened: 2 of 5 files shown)


src/watchmaker/workers/salt.py
```diff
@@ -266,6 +276,24 @@ def _build_salt_formula(self, extract_dir):
...
+            if self.salt_content_path:
+                self.log.debug(
+                    'Using salt content path: %s',
+                    self.salt_content_path
+                )
+                try:
+                    salt_content_src = os.sep.join((
+                        extract_dir, self.salt_content_path))
+                    salt_file_names = os.listdir(salt_content_src)
+                    for name in salt_file_names:
```
lorengordon, line 288: after we're done here, we can also remove the `salt_content_src` directory...

```suggestion
                    for name in salt_file_names:
                        srcname = os.sep.join((salt_content_src, name))
                        shutil.move(srcname, extract_dir)

                    shutil.rmtree(salt_content_src)
```

src/watchmaker/workers/salt.py
```diff
@@ -266,6 +276,24 @@ def _build_salt_formula(self, extract_dir):
                 filepath=salt_content_file,
                 to_directory=extract_dir
             )
+            if self.salt_content_path:
+                self.log.debug(
+                    'Using salt content path: %s',
+                    self.salt_content_path
+                )
+                try:
+                    salt_content_src = os.sep.join((
```
lorengordon, line 285: I think this line can be moved out of the try block...

```suggestion
                salt_content_src = os.sep.join((
                	extract_dir, self.salt_content_path))
                try:
                    salt_file_names = os.listdir(salt_content_src)
                    ...
```

tests/test_saltworker.py
```diff
@@ -355,6 +355,49 @@ def test_linux_salt_content_none(
...
+    """Test that Pythonic None can be used without error rather than 'None'."""
+    # setup ========================
+    system_params = {}
+    salt_config = {}
+    system_params["prepdir"] = "96003f32-5808-4ef8-a573-763b7f47ba9d"
+    system_params["logdir"] = "0585f9d7-ed0e-4a1b-ac0d-2b10a245a0eb"
+    system_params["workingdir"] = "35f411db-355b-4953-ae31-d6f592753e58"
+
+    salt_config["salt_content"] = "d002be6e-645d-4f58-97c9-8335df0ff5e4.zip"
+    salt_config["salt_content_path"] = "05628e08-f1be-474d-8c12-5bb6517fc5f9"
```
lorengordon, line 379: The test name and description track, `test_linux_salt_content_path_none`, but we're not setting `salt_content_path` to `None` here? Maybe two tests are needed?

src/watchmaker/workers/salt.py
spavuluri: roger! will do that.

tests/test_saltworker.py
spavuluri: yeah, I had started to test the arg to `None` and then evolved to adding more test coverage. 2 separate tests make sense.

spavuluri:
@lorengordon this is ready for re-review with suggested changes.

src/watchmaker/workers/salt.py
```diff
@@ -262,10 +273,47 @@ def _build_salt_formula(self, extract_dir):
...
             )
+            if self.salt_content_path:
+                self.log.debug(
+                    'Using salt content path: %s',
+                    self.salt_content_path
+                )
+                salt_content_src = os.sep.join((
+                    temp_extract_dir, self.salt_content_path))
+                salt_files_dir = None
+                for (i, path) in enumerate(glob.glob(salt_content_src)):
```
lorengordon, line 294: any particular rationale for the logic here? seems like it could be a bit simpler...

```
salt_content_glob = glob.glob(salt_content_src)
if length(salt_content_glob) > 0:
    # ... raise
salt_files_dir = salt_content_glob[0]
```

src/watchmaker/workers/salt.py
```diff
@@ -262,10 +273,47 @@ def _build_salt_formula(self, extract_dir):
...
+            temp_extract_dir = extract_dir
+            if self.salt_content_path:
+                temp_extract_dir = os.sep.join((
+                    self.working_dir,
+                    'salt-archive'
+                ))
             self.extract_contents(
                 filepath=salt_content_file,
-                to_directory=extract_dir
+                to_directory=temp_extract_dir
```
lorengordon, line 284: i see we're now extracting to `temp_extract_dir` in all cases but only *IF* `salt_content_path` is set do we move content back to `extract_dir`. when `salt_content_path` is not set, the content is not moved back. i think case needs to be handled.

also, it seems with these changes the function definition is becoming a leaky abstraction. it seems what we care about is not necessarily that we are extracting into `extract_dir`. or is it? is another function moving the extracted files _again_? might need to take a look at how this function is called, what sets `extract_dir` and maybe rename the argument to account for the actual usage?

src/watchmaker/workers/salt.py
```diff
@@ -262,10 +273,47 @@ def _build_salt_formula(self, extract_dir):
...
+                                      self.salt_content)
+                            self.log.critical(msg)
+                            raise WatchmakerException(msg)
+                        salt_files_dir = path
+                self.log.debug('Using extracted content from: %s',
+                               salt_files_dir)
+
+                for subdir in next(os.walk(salt_files_dir))[1]:
+                    if (
+                            not subdir.startswith('.') and
```
lorengordon, line 309: this all seems indented a bit much...
```suggestion
                        not subdir.startswith('.') and
                        not os.path.exists(os.sep.join((extract_dir, subdir)))
```

src/watchmaker/workers/salt.py
```diff
@@ -262,10 +273,47 @@ def _build_salt_formula(self, extract_dir):
...
+                                  ' \'{0}\' in {1}'.format(
+                                      self.salt_content_path,
+                                      self.salt_content)
+                            self.log.critical(msg)
+                            raise WatchmakerException(msg)
+                        salt_files_dir = path
+                self.log.debug('Using extracted content from: %s',
+                               salt_files_dir)
+
+                for subdir in next(os.walk(salt_files_dir))[1]:
```
lorengordon, line 307: this logic looks familiar, but the diff doesn't show it being moved or anything... is this copied from elsewhere in watchmaker?

src/watchmaker/workers/salt.py
```diff
@@ -40,6 +41,13 @@ class SaltBase(WorkerBase, PlatformManagerBase):
             - *Linux*: ``/srv/watchmaker/salt``
             - *Windows*: ``C:\Watchmaker\Salt\srv``
 
+        salt_content_path: (:obj:`str`)
+            Used in conjunction with the "salt_content" arg.
+            Path to the salt content files inside the provided salt_content
```
lorengordon, line 46: Would be helpful to note the glob support in the docs...

src/watchmaker/workers/salt.py
spavuluri: yeah, the added logic is there to only raise error when the glob matches multiple directories, not files necessarily. Too much of an edge case, you think?

src/watchmaker/workers/salt.py
spavuluri: So, we start out with setting `temp_extract_dir = extract_dir` on line 276. That should handle the case you mention. 
`extract_dir` is essentially the `salt_srv` dir. In Linux, for e.g. this is '/srv/watchmaker/salt' where the salt files are extracted to. 
https://github.com/plus3it/watchmaker/blob/62400d9b27f6beac5a4af5d89daf7c072429ce87/src/watchmaker/workers/salt.py#L683
The archive is currently extracted directly into that dir. What this abstraction does, is in case where `salt_content_path` is set, the archive is extracted to the working directory, then from the glob are copied into the `salt_srv` a.k.a the `extract_dir` dir.

src/watchmaker/workers/salt.py
spavuluri: Yup, that indent needs to go.

src/watchmaker/workers/salt.py
spavuluri: yup, when bundled content is used, files are copied over after a check to see if they already exist. I can add a util method to facilitate re-use.
https://github.com/plus3it/watchmaker/blob/62400d9b27f6beac5a4af5d89daf7c072429ce87/src/watchmaker/workers/salt.py#L273

src/watchmaker/workers/salt.py
spavuluri: Will do. Missed updating the doc.

src/watchmaker/workers/salt.py
lorengordon: i would say yes, too much an edge case. user can just use a more specific glob, right?

src/watchmaker/workers/salt.py
lorengordon: > So, we start out with setting `temp_extract_dir = extract_dir` on line 276. That should handle the case you mention.

oops i missed that assignment. you're right! that pretty much addresses the entire prior comment. 😂  ok, so the only thing i don't like about this approach is that we test `if self.salt_content_path` twice. what if we reverse the logic here and then return immediately?

```suggestion
            if not self.salt_content_path:
                self.extract_contents(
                    filepath=salt_content_file,
                    to_directory=extract_dir,
                )
                return
            
            # now we can proceed with `salt_content_path` logic without any condition
```

src/watchmaker/utils/__init__.py
```diff
@@ -119,3 +119,20 @@ def clean_none(value):
         return None
 
     return value
+
+
+def copy_subdirectories(src_dir, dest_dir, log):
```
lorengordon, line 124: make `log` optional, otherwise the `if log` test doesn't make much sense
```suggestion
def copy_subdirectories(src_dir, dest_dir, log=None):
```

src/watchmaker/workers/salt.py
```diff
@@ -273,64 +276,43 @@ def _build_salt_formula(self, extract_dir):
...
+                salt_content_glob = glob.glob(salt_content_src)
+                self.log.debug('salt_content_glob: %s', salt_content_glob)
+                if len(salt_content_glob) > 1:
+                    msg = 'Found multiple paths matching' \
+                          ' \'{0}\' in {1}'.format(
+                              self.salt_content_path,
+                              self.salt_content)
+                    self.log.critical(msg)
+                    raise WatchmakerException(msg)
+                salt_files_dir = salt_content_glob[0]
```
lorengordon, line 306: guessing if salt_content_path doesn't match anything, we get an empty list and an indexerror?

```suggestion
                try:
                    salt_files_dir = salt_content_glob[0]
                except IndexError:
                    ... log / raise msg about bad salt_content_path
```

src/watchmaker/workers/salt.py
```diff
@@ -273,64 +276,43 @@ def _build_salt_formula(self, extract_dir):
...
+
+                watchmaker.utils.copy_subdirectories(
```
lorengordon, line 308: after this, we still have `temp_extract_dir` hanging around right? make sense to remove it?

src/watchmaker/workers/salt.py
spavuluri: yup.

src/watchmaker/workers/salt.py
spavuluri: Thought about doing that, but the `temp_extract_dir` is being created within the `working_dir`. And since that gets removed during the cleanup routine, figured I'd not bother removing it.
https://github.com/plus3it/watchmaker/blob/62400d9b27f6beac5a4af5d89daf7c072429ce87/src/watchmaker/managers/platform.py#L242

spavuluri: let me know if that's not a good reason to leave it in there though.

lorengordon: no, that makes sense to me. can address later if we realize otherwise...

spavuluri:
@lorengordon that last set of updates is done.

lorengordon:
go codebuild go

PR #1214, opened 2020-04-28
Add pip install to salt worker

YakDriver:


YakDriver:
go codebuild go

YakDriver:
I switched the config to one that has a pip install section for this test (`config-pip.yaml`). When it completes, I'll switch back.

YakDriver:
go codebuild go

lorengordon:
you can always just run the job through the console, overriding the options just for that build...

src/watchmaker/__init__.py
```diff
@@ -152,6 +152,14 @@ class Arguments(dict):
 
                 ou_path="OU=Super Cool App,DC=example,DC=com"
 
+        pip_install: (:obj:`str`)
```
lorengordon, line 155: I don't think we need to update the cli with this arg at the moment. Cli args come in as strings, and we are expecting this to be a list, so this won't work on the cli anyway. And that's fine for now, as this is most likely to be managed solely through the config file.

src/watchmaker/__init__.py
```diff
@@ -194,6 +202,8 @@ def __init__(
             kwargs.pop('environment', None) or Arguments.DEFAULT_VALUE)
         self.salt_states = watchmaker.utils.clean_none(
             kwargs.pop('salt_states', None) or Arguments.DEFAULT_VALUE)
+        self.pip_install = watchmaker.utils.clean_none(
```
lorengordon, line 205: See above, can remove this from here for now.

src/watchmaker/workers/salt.py
```diff
@@ -84,6 +84,10 @@ class SaltBase(WorkerBase, PlatformManagerBase):
             E.g. ``"OU=SuperCoolApp,DC=example,DC=com"``
             (*Default*: ``''``)
 
+        pip_install: (:obj:`str`)
```
lorengordon, line 87:
```suggestion
        pip_install: (:obj:`list`)
```

src/watchmaker/workers/salt.py
```diff
@@ -100,6 +104,7 @@ def __init__(self, *args, **kwargs):
         self.ou_path = kwargs.pop('ou_path', None) or ''
         self.admin_groups = kwargs.pop('admin_groups', None) or ''
         self.admin_users = kwargs.pop('admin_users', None) or ''
+        self.pip_install = kwargs.pop('pip_install', None) or ''
```
lorengordon, line 107:
```suggestion
        self.pip_install = kwargs.pop('pip_install', None) or []
```

src/watchmaker/workers/salt.py
```diff
@@ -84,6 +84,10 @@ class SaltBase(WorkerBase, PlatformManagerBase):
             E.g. ``"OU=SuperCoolApp,DC=example,DC=com"``
             (*Default*: ``''``)
 
+        pip_install: (:obj:`str`)
+            Python modules to be installed prior to applying the high state.
+            (*Default*: ``None``)
```
lorengordon, line 89:
```suggestion
            (*Default*: ``[]``)
```

src/watchmaker/workers/salt.py
```diff
@@ -588,6 +593,7 @@ def __init__(self, *args, **kwargs):
             'policycoreutils-python',
             'selinux-policy-targeted',
             'salt-minion',
+            'python-pip'
```
lorengordon, line 596:
```suggestion
            'python-pip',
```

lorengordon: One complicating factor is that `python-pip` is for python2. This will work when installing the python2 versions of salt. It will not work when installing the python3 versions of salt. And, it will install an entire python2 system-level runtime if one is not installed (such as on EL8), which would be way undesirable if using python3 salt.

I don't have another suggestion just yet, but this needs some noodling...

src/watchmaker/workers/salt.py
```diff
@@ -661,6 +667,49 @@ def _set_grain(self, grain, value):
         self.log.info('Setting grain `%s` ...', grain)
         super(SaltLinux, self)._set_grain(grain, value)
 
+    def _install_pip_modules(self):
+        if self.pip_install:
+            python = self.run_salt(['grains.get', 'pythonexecutable'])
+            py_exec = python['stdout'].decode().split('\n')[1].strip()
+            self.log.info('Salt Python interpreter: `%s`', py_exec)
+
+            try:
```
lorengordon, line 676: Let's extract the stuff about installing pip to its own function

src/watchmaker/workers/salt.py
```diff
@@ -661,6 +667,49 @@ def _set_grain(self, grain, value):
...
+                get_pip = [
+                    'curl',
+                    'https://bootstrap.pypa.io/get-pip.py',
+                    '-o',
+                    'get-pip.py'
+                ]
+                self.call_process(get_pip)
+                self.call_process([py_exec, 'get-pip.py'])
+
+            for module in self.pip_install:
```
lorengordon, line 693: One the bits about installing pip itself are separate, both linux and windows ought to be able to use the same logic to install packages with pip, so these bits can move into the SaltWorker instead of the Linux/Windows workers

lorengordon: Pip can install multiple modules just fine, so no need to loop like this

src/watchmaker/workers/salt.py
```diff
@@ -661,6 +667,49 @@ def _set_grain(self, grain, value):
...
+                ensure = self.call_process([py_exec, '-m', 'ensurepip'])
+                self.log.info('Ensure pip: %s', ensure['stdout'])
+                ver = self.call_process([py_exec, '-m', 'pip', '--version'])
+                self.log.info('Pip version: %s', ver['stdout'])
+            except WatchmakerException:
+                self.log.info('Pip not installed for Salt interpreter!')
+                self.log.info('Attempting pip install...')
+                get_pip = [
+                    'curl',
+                    'https://bootstrap.pypa.io/get-pip.py',
```
lorengordon, line 686: This won't work on other networks. It will need to be a new config value that can _default_ to this (or for better compatibility, to the version that works with py26).

src/watchmaker/workers/salt.py
```diff
@@ -661,6 +667,49 @@ def _set_grain(self, grain, value):
...
+                    '-o',
+                    'get-pip.py'
```
lorengordon, line 688: where does this file end up? we should make sure we clean up.

src/watchmaker/workers/salt.py
```diff
@@ -661,6 +667,49 @@ def _set_grain(self, grain, value):
...
+
+            for module in self.pip_install:
+                self.log.info('Pip installing module: `%s`', module)
+                pip_cmd = [
+                    py_exec,
+                    '-m',
+                    'pip',
+                    'install',
+                    module,
+                    '--ignore-installed'
```
lorengordon, line 701: Why add this option? Also, systems that do not have connectivity to the public pypi will need to be able to pass their own options, at least to set the pypi index server. Probably need to consider another config value for user-specified pip options.

src/watchmaker/workers/salt.py
```diff
@@ -834,6 +896,7 @@ def install(self):
         self.log.info('Refreshing package database...')
         self.run_salt('pkg.refresh_db')
 
+        self._install_pip_modules()
```
lorengordon, line 899: would prefer to do this in the same place on windows as linux, just after `self._install_package()`, unless there is a good reason not to.

src/watchmaker/workers/salt.py
lorengordon: I'm not sure `self.call_process()` can handle it, but it may be cleaner to pipe curl to python... i.e. `curl https://<get-pip.py> | <py_exec> -`

lorengordon: Hmm, this might still get caught up in environment issues. See how we were setting the index url for the get-pip.py execution previously (via userdata)... https://github.com/plus3it/watchmaker/pull/843/files

src/watchmaker/workers/salt.py
YakDriver: The system interpreter has pure distutils installed packages that make pip fail when installing. For example, requests is distutils installed and is a prerequisite of hvac so it fails without this option.

src/watchmaker/workers/salt.py
YakDriver: Can't move to SaltWorker since Linux/Windows are distinct (e.g., Windows is `self.run_salt()` vs. Linux is `self.call_process()`, Windows is comma separated vs. Linux is space separated)

src/watchmaker/workers/salt.py
YakDriver: Using vendored file

src/watchmaker/workers/salt.py
lorengordon: No, can use the same `call_process()` option on windows. Using salt with it's `pip.install` function _does not_ actually install to salt's python runtime anyway, you have to also pass the python script location to get it installed in the right location.

YakDriver: argh - didn't realize. better to know now than when trying to run the formula!

YakDriver: this is fixed

YakDriver:
go codebuild go

.travis.yml
```diff
@@ -89,7 +89,7 @@ jobs:
       if: type != pull_request AND (branch in (master, develop) OR tag is present)
       python: 3.6
       install:
-        - echo "Install"
+        - git submodule update --init --recursive
```
lorengordon, line 92: did you actually see a problem requiring this? travis generally does this by default...

docs/configuration.md
```diff
@@ -107,6 +107,29 @@ Parameters supported by the Salt Worker:
...
+-   `pip_args` (_list_): Options to be passed to pip when installing package(s). Options with values should be passed with the option and value as separate list items.
+
+    Linux example where emoji is a value that corresponds to the `--progress-bar` option:
+
+    ```yaml
+    linux:
+      - salt:
+          pip_args:
+            - --ignore-installed
+            - --progress-bar
```
lorengordon, line 127: Might be more clear to just use the --arg=value syntax...
```suggestion
            - --progress-bar=emoji
```

src/watchmaker/workers/salt.py
```diff
@@ -360,6 +381,97 @@ def _get_failed_states(self, state_ret):
...
+
+        ver = self.call_process([self.py_exec, '-m', 'pip', '--version'])
+        self.log.info('Pip version: %s', ver['stdout'])
+
+    def _install_pip_packages(self):
+        if not self.pip_install:
+            self.log.info('No Python packages specified for install in config')
+            return 0
+
+        self.py_exec = self._get_grain('pythonexecutable')
```
lorengordon, line 435: I think things will flow a little better if `py_exec` is just a required attribute on `_install_pip()` and `_upgrade_pip()`... Can then remove it as a class attribute.

```suggestion
        py_exec = self._get_grain('pythonexecutable')
```

src/watchmaker/workers/salt.py
```diff
@@ -360,6 +381,97 @@ def _get_failed_states(self, state_ret):
...
+            "-m",
+            "pip",
+            "install",
+            "--upgrade",
+            "pip",
+            "--index-url",
+            self.pip_index
+        ]
+        try:
+            self.call_process(cmd)
```
lorengordon, line 422: We can just use the `raise_error` argument to disable the exception...
```suggestion
            self.call_process(cmd, raise_error=False)
```

src/watchmaker/workers/salt.py
```diff
@@ -360,6 +381,97 @@ def _get_failed_states(self, state_ret):
...
+            'Pip installing module(s): `%s`', " ".join(self.pip_install))
+        pip_cmd = [
+            self.py_exec,
+            '-m',
+            'pip',
+            'install'
+        ]
+        pip_cmd.extend(self.pip_install)
+        pip_cmd.extend(['--index-url', self.pip_index])
+        if self.pip_args:
```
lorengordon, line 461: We don't really need this test. `pip_args` must be a list, and extend does the right thing with an empty list.

src/watchmaker/workers/salt.py
```diff
@@ -360,6 +381,97 @@ def _get_failed_states(self, state_ret):
...
+            self.call_process(cmd)
+        except WatchmakerException:
+            self.log.info(
+                'Unable to upgrade pip, continuing with legacy version')
+
+        ver = self.call_process([self.py_exec, '-m', 'pip', '--version'])
+        self.log.info('Pip version: %s', ver['stdout'])
+
+    def _install_pip_packages(self):
+        if not self.pip_install:
```
lorengordon, line 431: Makes more sense to do this test in the `install()` routine...

```
if self.pip_install:
  self._install_pip_packages()
```

src/watchmaker/workers/salt.py
```diff
@@ -360,6 +381,97 @@ def _get_failed_states(self, state_ret):
...
+        self.call_process(pip_cmd)
+
+        pip_list = [
+            self.py_exec,
+            '-m',
+            'pip',
+            'list'
+        ]
+        mod_list = self.call_process(pip_list)
+        self.log.info('Installed Python packages: %s', mod_list['stdout'])
```
lorengordon, line 472: This isn't entirely an accurate statement... `pip list` will list _all_ pkgs it is aware of, not just what we asked it to install... 

But, all these subprocess execs are expensive. The commands are logged, the command output is logged. Errors are raised. We don't need this list or the return statement.

src/watchmaker/workers/salt.py
```diff
@@ -625,7 +737,7 @@ def __init__(self, *args, **kwargs):
         self.yum_pkgs = [
             'policycoreutils-python',
             'selinux-policy-targeted',
-            'salt-minion',
+            'salt-minion'
```
lorengordon, line 740: prefer to keep the trailing comma

src/watchmaker/workers/salt.py
```diff
@@ -360,6 +381,97 @@ def _get_failed_states(self, state_ret):
...
+        ver = self.call_process([self.py_exec, '-m', 'pip', '--version'])
+        self.log.info('Pip version: %s', ver['stdout'])
+
+    def _install_pip_packages(self):
+        if not self.pip_install:
+            self.log.info('No Python packages specified for install in config')
+            return 0
+
+        self.py_exec = self._get_grain('pythonexecutable')
+        self.log.info('Salt Python interpreter: `%s`', self.py_exec)
```
lorengordon, line 436:
```suggestion
        self.log.debug('Salt Python interpreter: `%s`', self.py_exec)
```

src/watchmaker/workers/salt.py
```diff
@@ -360,6 +381,97 @@ def _get_failed_states(self, state_ret):
...
+            self.pip_index
+        ]
+        try:
+            self.call_process(cmd)
+        except WatchmakerException:
+            self.log.info(
+                'Unable to upgrade pip, continuing with legacy version')
+
+        ver = self.call_process([self.py_exec, '-m', 'pip', '--version'])
+        self.log.info('Pip version: %s', ver['stdout'])
```
lorengordon, line 428:
```suggestion
        self.log.debug('Pip version: %s', ver['stdout'])
```

src/watchmaker/workers/salt.py
```diff
@@ -360,6 +381,97 @@ def _get_failed_states(self, state_ret):
...
+            "--index-url",
+            self.pip_index
+        ]
+        self.call_process(cmd)
+
+    def _upgrade_pip(self):
+        if not self.py_exec:
+            raise WatchmakerException("Python executable not set")
+
+        self.log.info('Attempting to upgrade pip...')
```
lorengordon, line 410:
```suggestion
        self.log.debug('Attempting to upgrade pip...')
```

src/watchmaker/workers/salt.py
```diff
@@ -360,6 +381,97 @@ def _get_failed_states(self, state_ret):
...
+        self.py_exec = self._get_grain('pythonexecutable')
+        self.log.info('Salt Python interpreter: `%s`', self.py_exec)
+
+        try:
+            ver = self.call_process([
+                self.py_exec,
+                '-m',
+                'pip',
+                '--version'])
+            self.log.info('Pip version: %s', ver['stdout'])
```
lorengordon, line 444:
```suggestion
            self.log.deubg('Pip version: %s', ver['stdout'])
```

src/watchmaker/workers/salt.py
```diff
@@ -360,6 +381,97 @@ def _get_failed_states(self, state_ret):
...
+        except WatchmakerException:
+            self.log.info('Pip not installed for Salt interpreter!')
```
lorengordon, line 446:
```suggestion
            self.log.debug('Pip not installed for Salt interpreter!')
```

src/watchmaker/workers/salt.py
```diff
@@ -360,6 +381,97 @@ def _get_failed_states(self, state_ret):
...
+            self._install_pip()
+
+        self._upgrade_pip()
+
+        self.log.info(
```
lorengordon, line 451: This one makes sense as `info`, since it is directly related to something the user asked us to do....

.travis.yml
YakDriver: overcorrected when trying to resolve a related issue

YakDriver:
I finished fixes, performed testing with `config-pip.yaml`. Let me know if there is anything else.

YakDriver:
go codebuild go

src/watchmaker/workers/salt.py
```diff
@@ -360,6 +380,78 @@ def _get_failed_states(self, state_ret):
             failed_states = state_ret
         return failed_states
 
+    def _install_pip(self, py_exec):
+        if not py_exec:
```
lorengordon, line 384: Can remove this check now, since it is a required argument

src/watchmaker/workers/salt.py
```diff
@@ -360,6 +380,78 @@ def _get_failed_states(self, state_ret):
...
+        cmd = [
+            py_exec,
+            get_pip,
+            "--index-url",
+            self.pip_index
+        ]
+        self.call_process(cmd)
+
+    def _upgrade_pip(self, py_exec):
+        if not py_exec:
```
lorengordon, line 406: remove

src/watchmaker/workers/salt.py
YakDriver: you can still pass in `py_exec=None` but i'm fine to remove

lorengordon: Yeah, but then we're in real, intentional user error territory. Can pass _any_ bad value for py_exec (e.g. non-empty, non-string, invalid path to python.exe) to generate all sorts of fun exceptions.

PR #1226, opened 2020-05-06
Prepare watchmaker 0.20.0 release

YakDriver:


YakDriver:
go codebuild go

YakDriver:
go codebuild go

YakDriver:
I had it as 0.20.0 and then thought you could think of it as a fix to Salt and switched it.

PR #1337, opened 2020-07-16
Updates default watchmaker config to use salt 2019.2.5

lorengordon:
Also setup a new webhook so the codebuild trigger will start two jobs, one for the custom content and one for the default content.


lorengordon:
go codebuild go

lorengordon:
go codebuild go

lorengordon:
go codebuild go

PR #1467, opened 2020-12-02
Ensures standalone build uses pinned versions of build deps

lorengordon:


lorengordon:
go codebuild go

lorengordon:
go codebuild go

lorengordon:
go codebuild go

lorengordon:
go codebuild go

lorengordon:
go codebuild go

ci/build.sh
```diff
@@ -2,16 +2,24 @@
...
 
-virtualenv $VIRTUAL_ENV_DIR
+python3 -m ensurepip --upgrade --default-pip
+python3 -m pip install -r requirements/pip.txt
+python3 -m pip --version
+
+python3 -m pip install -r requirements/basics.txt
+python3 -m pip list
+
+virtualenv --python=python3 $VIRTUAL_ENV_DIR
```
eemperor, line 12: Just curious.  Is this to add support for Python3 in the standalone?

lorengordon: the standalone has _always_ used python3. this is just being very very explicit because the docker container at the moment is running centos6, and so the default python is 2.6, which we definitely do not want to call by accident. :)

eemperor: Gotcha.  💡

eemperor (approved):
🚀

PR #1579, opened 2021-03-09
Prepares to use "main" as single long-lived branch

lorengordon:
* Merges develop to main
* Updates CI configs to operate on "main" branch


eemperor (approved):
The updated logic for the deployments looks fine to me.

PR #1597, opened 2021-03-22
Copies windows bootstrap to gravitbee staging dir

lorengordon:
This way the bootstrap will be rehosted in the deploy step along
with the standalone build files


eemperor (approved):
🚢

PR #1673, opened 2021-06-22
Allows user to pass yaml/json strings to extra args on cli

lorengordon:
Watchmaker passes all "extra" cli arguments to all workers, but previously,
users could only pass simple string arguments. Complex worker args could
be passed only in the config file.

This patch allows the user to pass them on the cli as json or yaml, and
deserializes them to python. E.g.

```
watchmaker --user-formulas '{"some-formula": "https:/url.zip"}'
```

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -173,7 +173,8 @@ Parameters supported by the Salt Worker:
 -   `salt_version` (_string_): (Linux-only) A git reference present in
     `git_repo`, such as a commit or a tag. If not specifid, the HEAD of the
     default branch will be used.
-    `installer_url` (_string_): (Windows-only) URL to the Salt Minion installer
+
+-   `installer_url` (_string_): (Windows-only) URL to the Salt Minion installer
     for Windows.
 
 ### yum worker (linux-only)
--- a/src/watchmaker/__init__.py
+++ b/src/watchmaker/__init__.py
@@ -261,7 +261,7 @@ class Client(object):
         ))
         # Set self.worker_args, removing `None` values from worker_args
         self.worker_args = dict(
-            (k, v) for k, v in worker_args.items()
+            (k, yaml.safe_load(v)) for k, v in worker_args.items()
             if v != Arguments.DEFAULT_VALUE
         )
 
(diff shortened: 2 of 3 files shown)


lorengordon:
go codebuild go

eemperor (commented):
Is `--user-formulas` the only arg that can make use of this now?  I guess the idea is support expanding capabilities of other args?

lorengordon:
TLDR; No, it applies to *every* "extra argument" supported by any worker.

Watchmaker has always passed any CLI "extra" argument to all workers specified in the config. An "extra" argument is one *not* specified as a [CLI option](https://github.com/plus3it/watchmaker/blob/635d01d6af722a7d46a6cd061e09eeb41d87f87c/src/watchmaker/cli.py#L37-L96). CLI options are output in `watchmaker --help`. The two workers, salt and yum, define their own arguments:

* https://watchmaker.cloudarmor.io/en/stable/api.html#module-watchmaker.workers.salt
* https://watchmaker.cloudarmor.io/en/stable/api.html#module-watchmaker.workers.yum

So you could *always* pass `--user-formulas` or *any other argument processed by the worker*, and watchmaker passes it through to the worker. It strips off leading hyphens and converting other hyphens to underscore, e.g. `user_formulas`, but otherwise didn't touch the input value.

The problem was that CLI values are input as plain strings, but some of the worker arguments, like `user_formulas`, are lists, or maps, or lists of maps, etc... So the worker would basically explode on the invalid type.

This patch uses the yaml library to "load" the string, which converts a serialized json or yaml string to the python equivalent. This way the worker gets an actual python object instead of just the string.

eemperor:
Ah.  If I'm understanding it correctly, this update would enable the ability to pass arguments that can function like the full `config.yaml` file correct?

lorengordon:
> Ah. If I'm understanding it correctly, this update would enable the ability to pass arguments that can function like the full `config.yaml` file correct?

I hesitate to say you could pass "any" setting from `config.yaml` this way, because it's only args that are passed to the workers. Now, at the moment, the only config setting that *isn't* a worker arg is `watchmaker_version`. So, "pretty much yes" is not entirely inaccurate. But it's not a guarantee, we may add more non-worker settings...

Although at that point, this qualification and special treatment of some args vs others will really start to itch, and we may look at a radical overhaul of the config format entirely (and maybe the cli, also).

eemperor (approved):
🚀 Lets give it go!

PR #1701, opened 2021-07-19
Use lowercase only when checking highstate

eemperor:


eemperor:
go codebuild go

src/watchmaker/workers/salt.py
```diff
@@ -631,20 +632,22 @@ def process_states(self, states, exclude):
             return
 
         cmds = []
-        states = states.lower().split(',')
```
lorengordon, line 634: This _should_ already be case insensitve... 

```
$ python
>>> 'highstate' in 'Highstate,foo,bar'.lower().split(',')
True
```

lorengordon: Oh I see, that _other_ state names are not case insensitive?

eemperor: > Oh I see, that _other_ state names are not case insensitive?

Yup.  That's the issue.

lorengordon:
```suggestion
        states = states.split(',')
```

lorengordon:
```suggestion
        states = states.split(',')
```

src/watchmaker/workers/salt.py
```diff
@@ -631,20 +632,22 @@ def process_states(self, states, exclude):
...
         salt_cmd = self.salt_state_args
 
-        if 'highstate' in states:
+        if 'highstate' in states.lower().split(','):
```
lorengordon, line 637: The split/split/join/join is really hard to understand. I think we can do a little better. Here's one idea.

```suggestion
        highstate = False
        for state in states:
            if state.lower() == 'highstate':
                highstate = True
                states.remove(state)

        if highstate:
            self.log.info('Applying the salt "highstate"')
            cmds.append(salt_cmd + ['state.highstate'])
```

src/watchmaker/workers/salt.py
```diff
@@ -631,20 +632,22 @@ def process_states(self, states, exclude):
...
+                states,
+                flags=re.IGNORECASE).split(',')).split())
             cmds.append(salt_cmd + ['state.highstate'])
 
         if states:
             self.log.info(
                 'Applying the user-defined list of states, states=%s',
                 states
             )
-            states = ','.join(states)
```
lorengordon, line 647: Adding this back depends on the approach above, but would be needed if going with my suggestion.

```suggestion
            states = ','.join(states)
```

lorengordon (commented):
Just offering another option, trying something that I hope is easier to understand than the split/split/join/join...

Would also like to get a test case for this, confirming that case is preserved for non-highstate input...

src/watchmaker/workers/salt.py
eemperor: Okay.  Let me give this suggestion a try.

eemperor: I updated to your suggestion.  Tests are passing as well.

lorengordon: I think it reads well. Would still like to see a test case demonstrating case preservations for the "extra" states...

eemperor:
> Just offering another option, trying something that I hope is easier to understand than the split/split/join/join...
> 
> Would also like to get a test case for this, confirming that case is preserved for non-highstate input...

I'll work on adding a test case.

eemperor:
@lorengordon I added the test case but it's failing when I try it locally in my working branch.  I'm not sure why the states are still being converted to lowercase in the test is failing.  Am I running it correctly?

```
[watchmaker]$ pytest tests/test_saltworker.py
_________________________________________________ test_process_states_multiple_states_case_sensitive _________________________________________________
tests/test_saltworker.py:173: in test_process_states_multiple_states_case_sensitive
    raise_error=False
/usr/lib64/python3.6/unittest/mock.py:814: in assert_called_with
    raise AssertionError(_error_message()) from cause
E   AssertionError: Expected call: mock(['--log-file', 'salt_call.debug.log', '--log-file-level', 'debug', '--log-level', 'error', '--out', 'quiet', '--return', 'local', 'state.sls', 'Foo,bAR'], log_pipe='stderr', raise_error=False)
E   Actual call: mock(['--log-file', 'salt_call.debug.log', '--log-file-level', 'debug', '--log-level', 'error', '--out', 'quiet', '--return', 'local', 'state.sls', 'foo,bar'], log_pipe='stderr', raise_error=False)
```

lorengordon:
>I added the test case but it's failing when I try it locally in my working branch. I'm not sure why the states are still being converted to lowercase and the test fails. Am I running it correctly?

@eemperor Most likely, you installed watchmaker from PyPi instead from source. Pytest imports watchmaker, so it gets the library you have installed. Try `python -m pip uninstall watchmaker` and then `python -m pip install -e .` (while in the watchmaker directory, with this branch checked out).

lorengordon:
And if you have multiple pythons on your system, you'll also want to make sure pytest and watchmaker are installed in the same python runtime...

lorengordon (approved):
🚢

eemperor:
> > I added the test case but it's failing when I try it locally in my working branch. I'm not sure why the states are still being converted to lowercase and the test fails. Am I running it correctly?
> 
> @eemperor Most likely, you installed watchmaker from PyPi instead from source. Pytest imports watchmaker, so it gets the library you have installed. Try `python -m pip uninstall watchmaker` and then `python -m pip install -e .` (while in the watchmaker directory, with this branch checked out).

This was it.  I ran the uninstall and installed from source.  Works now!

PR #1709, opened 2021-07-26
Defaults --salt-states to `None` on cli

lorengordon:


lorengordon:
go codebuild go

eemperor (approved):
🚢

I tested with config files with `salt_states` set to multiple states and also with `null` to verify the handling of multiple states or when nothing is provided.

PR #1711, opened 2021-07-28
Allows system-level configs to override the same option in the "all" config

lorengordon:


lorengordon:
go codebuild go

src/watchmaker/__init__.py
```diff
@@ -366,8 +374,11 @@ def _get_config(self):
                     config[worker_name] = {'config': worker_config}
                     self.log.debug('%s config: %s', worker_name, worker_config)
                 else:
-                    # Worker present in both config_system and config_all
-                    config[worker_name]['config'].update(worker_config)
+                    # Worker is present in both config_system and config_all,
+                    # config[worker_name]['config'] is from config_system,
+                    # worker_config is from config_all
+                    worker_config.update(config[worker_name]['config'])
+                    config[worker_name]['config'] = worker_config
```
eemperor, line 381: I wanted to confirm I'm understanding this update correctly.  When the same worker exists in both config_system and config_all, config_system worker configs will over-write the config_all worker configs?

lorengordon: Yes.

Due to the loop order, we get into this logic branch only when the same worker is present in both config sections. At this point, `config[worker_name]['config']` is always `config_system`, and `worker_config` is always `config_all`.

Before, it was updating the config_system values with the config_all values:

```
config[worker_name]['config'].update(worker_config)
```

This patch basically just reverses the prior order of operations. It first updates the config_all values (`worker_config`) with the config_system values (`config[worker_name]['config']`), and then replaces the config with the new dictionary.

(And yeah, this whole loop is confusing and probably needs a rewrite, but this was the least amount of change I could think of that would reverse this precedence...)

eemperor: Yeah.  Took a bit to wrap my head around the dictionary construction/manipulation.  I'm good now.

🚢

eemperor (approved):

PR #1715, opened 2021-07-29
Outputs a more helpful message when a cli extra_argument is invalid YAML/JSON

lorengordon:
Example:

```
> watchmaker --salt-states '*RHEL' --log-dir /tmp/log/watchmaker
...
Traceback (most recent call last):
  File "/home/loren/.local/bin/watchmaker", line 33, in <module>
    sys.exit(load_entry_point('watchmaker', 'console_scripts', 'watchmaker')())
  File "/usr/lib/python3/dist-packages/click/core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "/usr/lib/python3/dist-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/usr/lib/python3/dist-packages/click/core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/lib/python3/dist-packages/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "/home/loren/git/repos/watchmaker/src/watchmaker/cli.py", line 109, in main
    watchmaker_client = watchmaker.Client(watchmaker_arguments)
  File "/home/loren/git/repos/watchmaker/src/watchmaker/__init__.py", line 301, in __init__
    raise InvalidValue(msg)
watchmaker.exceptions.InvalidValue: Failed to parse argument value as YAML. Check the format and/or properly quote the value when using the CLI to account for shell interpolation. YAML error: found undefined alias 'RHEL'
  in "<unicode string>", line 1, column 1:
    *RHEL
    ^

```

instead of:

```
Traceback (most recent call last):
  File "/home/loren/git/repos/watchmaker/src/watchmaker/__init__.py", line 289, in __init__
    self.worker_args = dict(
  File "/home/loren/git/repos/watchmaker/src/watchmaker/__init__.py", line 290, in <genexpr>
    (k, yaml.safe_load(v)) for k, v in worker_args.items()
  File "/usr/lib/python3/dist-packages/yaml/__init__.py", line 162, in safe_load
    return load(stream, SafeLoader)
  File "/usr/lib/python3/dist-packages/yaml/__init__.py", line 114, in load
    return loader.get_single_data()
  File "/usr/lib/python3/dist-packages/yaml/constructor.py", line 49, in get_single_data
    node = self.get_single_node()
  File "/usr/lib/python3/dist-packages/yaml/composer.py", line 36, in get_single_node
    document = self.compose_document()
  File "/usr/lib/python3/dist-packages/yaml/composer.py", line 55, in compose_document
    node = self.compose_node(None, None)
  File "/usr/lib/python3/dist-packages/yaml/composer.py", line 68, in compose_node
    raise ComposerError(None, None, "found undefined alias %r"
yaml.composer.ComposerError: found undefined alias 'RHEL'
  in "<unicode string>", line 1, column 1:
    *RHEL
    ^
```

### DIFF
--- a/src/watchmaker/__init__.py
+++ b/src/watchmaker/__init__.py
@@ -19,7 +19,7 @@ from compatibleversion import check_version
 
 import watchmaker.utils
 from watchmaker import static
-from watchmaker.exceptions import WatchmakerException
+from watchmaker.exceptions import InvalidValue, WatchmakerException
 from watchmaker.logger import log_system_details
 from watchmaker.managers.worker_manager import (LinuxWorkersManager,
                                                 WindowsWorkersManager)
@@ -283,11 +283,23 @@ class Client(object):
             (k.lstrip('-').replace('-', '_'), v) for k, v in zip(
                 *[iter(extra_arguments)] * 2)
         ))
-        # Set self.worker_args, removing `None` values from worker_args
-        self.worker_args = dict(
-            (k, yaml.safe_load(v)) for k, v in worker_args.items()
-            if v != Arguments.DEFAULT_VALUE
-        )
+
+        try:
+            # Set self.worker_args, removing `None` values from worker_args
+            self.worker_args = dict(
+                (k, yaml.safe_load(v)) for k, v in worker_args.items()
+                if v != Arguments.DEFAULT_VALUE
+            )
+        except yaml.YAMLError as exc:
+            if hasattr(exc, "problem_mark"):
+                msg = (
+                    "Failed to parse argument value as YAML. Check the format "
+                    "and/or properly quote the value when using the CLI to "
+                    "account for shell interpolation. YAML error: {0}"
+                ).format(str(exc))
+                self.log.critical(msg)
+                raise InvalidValue(msg)
+            raise
 
         self.config = self._get_config()
 


src/watchmaker/__init__.py
```diff
@@ -283,11 +283,23 @@ def __init__(self, arguments):
...
+                (k, yaml.safe_load(v)) for k, v in worker_args.items()
+                if v != Arguments.DEFAULT_VALUE
+            )
+        except yaml.YAMLError as exc:
+            if hasattr(exc, "problem_mark"):
+                msg = (
+                    "Failed to parse argument value as YAML. Check the format "
+                    "and/or properly quote the value when using the CLI to "
+                    "account for shell interpolation. YAML error: {0}"
+                ).format(str(exc))
```
eemperor, line 299: Just a minor formatting comment.  Can you add newlines so the error message prints similar to how it's in the code?


Something like this?
```
raise InvalidValue(msg) watchmaker.exceptions.InvalidValue:
Failed to parse argument value as YAML. Check the format
and/or properly quote the value when using the CLI to
account for shell interpolation.
YAML error: found undefined alias 'RHEL'
in "<unicode string>", line 1, column 1:
    *RHEL
    ^
```

lorengordon: Eh, I'd rather not? The line breaks in the code are dictated by the indent of the code and the lint check for 80 chars. Trying to force those same line breaks on the shell, which is itself dependent on the terminal's line-wrapping logic, tends to result in unexpected pain... Improving how watchmaker manages its output for user consumption is a larger project. It does nothing now, just uses "raise" to bail. This kind of output handling would better be managed in a set of exception classes, instead of trying to format individual exception messages...

eemperor: Okay.  Yeah, handling is the outputs as a whole is better.  Good for me.

eemperor (approved):

PR #1737, opened 2021-08-30
Adds modified version to scap S3 path for non-tag deployments

eemperor:


.travis.yml
```diff
@@ -114,7 +115,7 @@ jobs:
             tags: true
         - provider: script
           # publish dev scap reports
-          script: bash ci/codebuild_start_wait.sh $CODEBUILD_PROJECT s3://$DEV_BUCKET/$S3_KEYFIX/$RELEASE_VERSION/scans
+          script: bash ci/codebuild_start_wait.sh $CODEBUILD_PROJECT s3://$DEV_BUCKET/$S3_KEYFIX/$CURRENT_VERSION/scans
```
lorengordon, line 118: For both cases, here and RELEASE_VERSION, I think instead we can just use the env provided by gravitybee...

```suggestion
          script: bash ci/codebuild_start_wait.sh $CODEBUILD_PROJECT s3://$DEV_BUCKET/$S3_KEYFIX/$GB_ENV_APP_VERSION/scans
```

lorengordon (changes requested):
Gravitybee gives us an env we can use instead...

.travis.yml
eemperor: Good call.  Switched both to `GB_ENV_APP_VERSION`.

lorengordon (approved):
🚢

PR #1751, opened 2021-09-10
Handles 'none' values passed on the cli properly

lorengordon:
Once upon a time, a `'none'` value on the cli was handled properly. It broke when we adding yaml parsing of cli values. Also, changing the `--salt-states` cli default to `None` broke how that setting was passed through to the salt worker, so it was always executing the salt worker default of `'highstate'`.

This patch adjusts the yaml parsing so the `None` value is parsed as `'null'` which again just returns `None`. This fixes the cli handling of `'none'` values.

This also adjusts the salt worker default for `salt_states` so `None` is honored when passed explicitly. When `salt_states` is not passed at all, the default value remains `'highstate'`.

Tests were updated or added to validate the correct behavior.

### DIFF
--- a/src/watchmaker/__init__.py
+++ b/src/watchmaker/__init__.py
@@ -287,7 +287,8 @@ class Client(object):
         try:
             # Set self.worker_args, removing `None` values from worker_args
             self.worker_args = dict(
-                (k, yaml.safe_load(v)) for k, v in worker_args.items()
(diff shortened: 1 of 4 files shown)


lorengordon:
go codebuild go

eemperor (approved):
🚢

PR #1755, opened 2021-09-13
Updates default config.yaml to use Salt 3003.3

eemperor:


eemperor:
go codebuild go

src/watchmaker/static/config.yaml
```diff
@@ -1,4 +1,4 @@
-watchmaker_version: ">= 0.21.9"
+watchmaker_version: ">= 0.23.2"
```
lorengordon, line 1: i don't think we need to change this... this is for when we modify config in a way that the parser of an earlier version won't recognize the format/type/syntax... in this case, an earlier version of watchmaker could use this config just fine.

eemperor: Yeah.  I was questioning whether it was needed but wasn't sure.  I'll undo the change.

lorengordon (approved):
:shipit:

PR #1805, opened 2021-10-18
Adds support for python 3.9 and 3.10

lorengordon:
Watchmaker has had this bug for a while now. CLIENT_AUTH is used by
servers to verify clients. SERVER_AUTH is used by clients to verify
servers.

Python 3.10 introduces PROTOCOL_TLS_CLIENT and PROTOCOL_TLS_SERVER,
which creates unique contexts for each scenario. Because Watchmaker
reversed the purpose for the context, it began to fail in Python 3.10.

For evidence, see the source for `create_default_context`:

* python/cpython@b494f59/Lib/ssl.py#L754-L760

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -180,8 +180,8 @@ specify something like this for EC2 userdata:
     $GitBranch = "<your-branch>"
 
     $BootstrapUrl = "https://watchmaker.cloudarmor.io/repo/releases/latest/watchmaker-bootstrap.ps1"
-    $PythonUrl = "https://www.python.org/ftp/python/3.6.8/python-3.6.8-amd64.exe"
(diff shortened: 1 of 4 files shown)


lorengordon:
go codebuild go

.travis.yml
```diff
@@ -16,10 +16,12 @@ python:
   - 3.6
   - 3.7
   - 3.8
+  - 3.9
+  - 3.10-dev
```
lorengordon, line 20: Unfortunately, travis-ci does not yet have a GA python 3.10 build. This works for now, I think. Tracking this post for updates: https://travis-ci.community/t/add-python-3-10/12220

eemperor (approved):
🚢

PR #2289, opened 2022-09-19
Add an EL8 entry for the yum config-map

### LINKED ISSUE #2288: [Feature] src/watchmaker/static/config.yaml needs an EL8 entry in the yum config-map
Currently, when running watchmaker with the default configuration, it attempts to match the distro and version using a config-map in the `src/watchmaker/static/config.yaml` file. However, that file lacks a suitable mapping. Will need to:
1. Add a suitable mapping
2. Ensure that the `salt-reposync-el8-python3.repo` actually exists in the default `url`-path (which it [already appears to](https://watchmaker.cloudarmor.io/yum.defs/saltstack/salt/3004.2/salt-reposync-el8-python3.repo))

ferricoxide:
Closes #2288

### DIFF
--- a/src/watchmaker/static/config.yaml
+++ b/src/watchmaker/static/config.yaml
@@ -37,6 +37,12 @@ linux:
             - centos
           el_version: 7
           url: https://watchmaker.cloudarmor.io/yum.defs/saltstack/salt/3004.2/salt-reposync-el7-python3.repo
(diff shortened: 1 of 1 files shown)


ferricoxide:
go codebuild go

eemperor (approved):

PR #2328, opened 2022-10-13
Create status section in the config

dwc0011:
This PR creates a status section in the config and processes the section.
It adds IMDS capability to detect AWS and Azure cloud providers.
If any of the status sections match the platform watchmaker is being run on it will tag the resource based on the configuration defined.  

This is still a work in process, I wanted to get an early review of the structure and have the CI exercised against the various supported python versions


src/watchmaker/status/providers/aws_status_provider.py
```diff
@@ -0,0 +1,82 @@
+# -*- coding: utf-8 -*-
+"""AWS Status Provider."""
+from __future__ import (absolute_import, division, print_function,
+                        unicode_literals, with_statement)
+
+import logging
+
+from watchmaker.status.providers.status_provider import (
```
lorengordon, line 8: Maybe?

```suggestion
from watchmaker.status.providers import (
```

src/watchmaker/utils/urllib/__init__.py
```diff
@@ -4,9 +4,10 @@
                         unicode_literals, with_statement)
 
 # pylint: disable=import-error
-from six.moves.urllib import error, parse, request  # noqa: F401
+from six.moves.urllib import error, parse, request  # noqa F401
 
-from watchmaker.utils.urllib.request_handlers import HAS_BOTO3, S3Handler
+from watchmaker.utils.conditions import HAS_BOTO3
```
lorengordon, line 9: I think we also talked about moving `conditions` to the top level:

```suggestion
from watchmaker.conditions import HAS_BOTO3
```

src/watchmaker/status/providers/aws_status_provider.py
dwc0011: On this one, the class in __init__.py is actually status and status_provider is an abstract class for the providers so I renamed to abstract.py instead

src/watchmaker/static/config.yaml
```diff
@@ -54,3 +55,14 @@ windows:
   - salt:
       salt_debug_log: null
       installer_url: https://watchmaker.cloudarmor.io/repo/saltstack/salt/windows/Salt-Minion-3004.2-1-Py3-AMD64-Setup.exe
+
+status:
+  targets:
+    - key: 'WatchmakerStatus'
+      required: False
+      target_type: 'aws'
+      status_type: 'Running'
```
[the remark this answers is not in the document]
dwc0011: I have left it for a placeholder for future options, I do use the status type running for now.  If we decide we do not need the status type then we can remove and refactor the usage pretty simply as it is all dicts.

src/watchmaker/status/__init__.py
```diff
@@ -0,0 +1,63 @@
...
+        """Get provider identifier."""
+        return self.identifier
+
+    def get_status_provider(self):
+        """Get provider."""
+        return self.status_provider
+
+    def tag_resource(self, status):
+        """Tag resource with key and status provided."""
+        if not self.status_provider or not self.targets:
```
lorengordon, line 46: Why would we call this function such that we would need to gate this?

src/watchmaker/status/providers/abstract.py
```diff
@@ -0,0 +1,32 @@
...
+"""Abstract Status Provider."""
+from __future__ import (absolute_import, division, print_function,
+                        unicode_literals, with_statement)
+
+import abc
+
+import six
+
+
+class StatusProviderError(Exception):
```
lorengordon, line 11: Thoughts on putting this in the `watchmaker.exceptions` module?

src/watchmaker/status/providers/aws.py
```diff
@@ -0,0 +1,83 @@
...
+        self.logger = logger or logging.getLogger(__name__)
+        self.metadata_id_url = (
+            'http://169.254.169.254/latest/meta-data/instance-id'
+        )
+        self.instance_id = None
+        self.initialize()
+
+    def initialize(self):
+        """Initialize instance id."""
+        if not self.instance_id:
```
lorengordon, line 29: I always like to reverse these conditions, where it helps reduce indentation.

```suggestion
        if self.instance_id:
            return

        try:
```

src/watchmaker/utils/imds/detect/__init__.py
```diff
@@ -0,0 +1,36 @@
...
+            log.debug("IMDS detected result is azure")
+            return "azure"
+
+    log.debug("IMDS detected result is unknown")
+    return "unknown"
+
+
+def is_excluded(identifier, excluded=None):
+    """Check if identifier is in excluded list."""
+    if not excluded:
```
lorengordon, line 34: That's an interesting way to handle the default of `None`, when the arg is a list. I usually end up normalizing the type with something like:

```suggestion
    excluded = excluded if excluded else []
```

src/watchmaker/utils/urllib/request_handlers.py
```diff
@@ -8,20 +8,13 @@
 
 from six.moves import urllib
 
-HAS_BOTO3 = False
-try:
-    import boto3
```
lorengordon, line 13: I _think_ this module might now be able to import boto3 unconditionally, because this module is only ever referenced by `watchmaker.utils.urllib` and that is already gated.

src/watchmaker/conditions/__init__.py
```diff
@@ -0,0 +1,23 @@
...
+    import boto3  # type: ignore # noqa: F401
+
+    HAS_BOTO3 = True
+except ImportError:
+    pass
+
+HAS_AZURE = False
+try:
+    from azure.core import pipeline  # type: ignore # noqa: F401
+    from azure.identity import (  # type: ignore # noqa: F401 E501
```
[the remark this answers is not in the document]
dwc0011: That was a holdover from before refactoring this code.

src/watchmaker/config/__init__.py
```diff
@@ -0,0 +1,129 @@
...
+            log.critical(msg)
+            raise
+
+    log.debug(
+        "Command-line arguments merged into worker configs: %s", worker_args
+    )
+
+    if not is_valid_status_config(config_status):
+        log.error("Status config is invalid %s", config_status)
+        config_status = None
```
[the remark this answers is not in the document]
dwc0011: Concur

src/watchmaker/config/status/__init__.py
```diff
@@ -0,0 +1,61 @@
...
+    "RUNNING": "Running",
+    "COMPLETE": "Completed",
+    "ERROR": "Error",
+}
+
+
+def is_valid_status_config(config):
+    """Validate config."""
+    is_valid = True
+    if config and "targets" in config:
```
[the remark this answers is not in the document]
dwc0011: Actually, this highlighted some other issues.  I was able to use this snippet to simplify things

src/watchmaker/config/status/__init__.py
```diff
@@ -0,0 +1,61 @@
...
+                    is_valid = False
+                    logging.error("Status target required value is not a bool")
+
+    return is_valid
+
+
+def get_status(status_key):
+    """Get formatted status message for status key provided."""
+    status = STATUS.get(status_key, None)
+    return status if status else status_key
```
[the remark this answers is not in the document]
dwc0011: I was allowing an option for custom messages or use pre-canned messages.

src/watchmaker/config/status/__init__.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+def is_target_required(target):
+    """Get whether target required."""
+    return target["required"]
+
+
+def get_targets_by_target_type(config_status, target_type):
+    """Get the targets for the target type."""
+    if config_status and target_type:
```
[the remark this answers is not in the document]
dwc0011: I moved the check to where the call is being made to this method.  But previously yes the identifier could have been None, not the config, so that was a wasted checked.  But after looking at it with fresh eyes I agree.

src/watchmaker/config/status/__init__.py
```diff
@@ -0,0 +1,61 @@
...
+        target_type = target_type.lower()
+        if target_type:
+            return [
+                target
+                for target in config_status["targets"]
+                if target["target_type"].lower() == target_type
+            ]
+
+    return None
```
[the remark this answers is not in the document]
dwc0011: Yes, after the previous comment changes, that is what this refactored to.

src/watchmaker/static/config.yaml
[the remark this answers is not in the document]
dwc0011: Yes

src/watchmaker/config/status/__init__.py
```diff
@@ -0,0 +1,61 @@
...
+import logging
+
+STATUS = {
+    "RUNNING": "Running",
+    "COMPLETE": "Completed",
+    "ERROR": "Error",
+}
+
+
+def is_valid_status_config(config):
```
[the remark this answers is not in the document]
dwc0011: concur

src/watchmaker/status/__init__.py
dwc0011: TL/DR;  you might not be on a valid provider, or for this provider you might not have a status target.  

If we returned the provider to watchmaker, every time we went to use the provider we would have to check if the provider was set, think of the Status class as a Proxy for tagging resources, it maintains what the provider and targets are for the platform/instance watchmaker is run on.  This instance does not have to be AWS or AZURE, so the provider would be None, or it may be AWS but there is no status config for this provider so the targets are none.

src/watchmaker/status/providers/abstract.py
dwc0011: Makes sense.

src/watchmaker/status/providers/aws.py
dwc0011: I can agree with that.

src/watchmaker/utils/imds/detect/__init__.py
dwc0011: Ill raise you one and say this can be done with this one liner instead return identifier in excluded if excluded else False

src/watchmaker/utils/urllib/request_handlers.py
dwc0011: I removed the check and import, added the import back.

dwc0011: Actually, we do need the try surrounding the import, not the HAS_BOTO3 line as that was removed to another spot, but without the try except it fails all unit tests

src/watchmaker/config/status/__init__.py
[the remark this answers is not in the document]
lorengordon: Can you elaborate how the custom messages work? This is just returning the key... but the key to what?

src/watchmaker/status/__init__.py
lorengordon: Does that mean we are calling this function even when the config would have nothing to actually tag?

src/watchmaker/utils/urllib/request_handlers.py
lorengordon: Hmm, can you check again exactly what was the error? While this works, I feel like maybe there should be a different fix, now that the condition is being handled elsewhere...

src/watchmaker/config/status/__init__.py
```diff
@@ -0,0 +1,62 @@
...
+
+    valid = True
+    for target in targets:
+        if "key" not in target or not target["key"]:
+            valid = False
+            logging.error("Status target is missing key or value")
+        if "target_type" not in target or not target["target_type"]:
+            valid = False
+            logging.error("Status target is missing target_type or value")
+        if "required" in target:
```
lorengordon, line 32: Can collapse the two if statements using get:

```suggestion
        if not isinstance(target.get("required"), bool):
```

src/watchmaker/static/config.yaml
```diff
@@ -54,3 +55,12 @@ windows:
...
       installer_url: https://watchmaker.cloudarmor.io/repo/saltstack/salt/windows/Salt-Minion-3004.2-1-Py3-AMD64-Setup.exe
+
+status:
+  targets:
+    - key: 'WatchmakerStatus'
+      required: False
+      target_type: 'aws'
+    - key: 'WatchmakerStatus'
+      required: False
+      target_type: 'azure'
```
lorengordon, line 66: Add an ending new line so github doesn't whine

src/watchmaker/config/status/__init__.py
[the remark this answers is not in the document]
dwc0011: If the key is a valid key in the STATUS dict
`STATUS = {
    "RUNNING": "Running",
    "COMPLETE": "Completed",
    "ERROR": "Error",
}`

it would return the value for that key as the status, otherwise it just returns what was pass in as the status.   
Essentially, the status is either a value from the map or what was passed in.  That would be addressed in the comments when I go through.  This was refactored a few times so I was holding off on that to document.

Is this necessary, no, but I thought what if we wanted a more elaborate message for one tag resource spot in the code, this would cover that use case without having another tag_resource method.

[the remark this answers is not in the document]
lorengordon: Ok, so sometimes the `status_key` is not a key, but the custom message? Like so?

```
get_status("RUNNING")
# "Running"

get_status("My custom status message here")
# "My custom status message here"
```

src/watchmaker/status/__init__.py
dwc0011: Yes.  We always call tag resource, but we don't always call provider.tag_resource. The proxy alleviate needing  conditionals wrapping the call or some declarative dict, which is essentially a conditional of a different means which the proxy is as well.

Could we check if config_status exists or not before calling, yes, that would be conditional above, but then we would still need to do the check inside of tag resource because it might not have targets for this provider, but it does have status configs.  Thats why I did all of the conditions in the call instead of every place I make the call

src/watchmaker/utils/urllib/request_handlers.py
dwc0011: `tests/test_utils.py:7: in <module>
    import watchmaker.utils
.tox/3.6-codecov/lib/python3.8/site-packages/watchmaker/__init__.py:18: in <module>
    import watchmaker.utils
.tox/3.6-codecov/lib/python3.8/site-packages/watchmaker/utils/__init__.py:13: in <module>
    from watchmaker.utils import urllib
.tox/3.6-codecov/lib/python3.8/site-packages/watchmaker/utils/urllib/__init__.py:10: in <module>
    from watchmaker.utils.urllib.request_handlers import S3Handler
.tox/3.6-codecov/lib/python3.8/site-packages/watchmaker/utils/urllib/request_handlers.py:10: in <module>
    import boto3  # type: ignore
E   ModuleNotFoundError: No module named 'boto3'`

The test discovery fails in tox as well as in vscode.

lorengordon: hmm, this stands out to me:

```
from watchmaker.utils.urllib.request_handlers import S3Handler
```

So I think then in `watchmaker.utils.urllib`, we can move the import into the gated check...

```
if HAS_BOTO3:
    from watchmaker.utils.urllib.request_handlers import S3Handler
    request.install_opener(request.build_opener(S3Handler))
```

src/watchmaker/status/__init__.py
```diff
@@ -0,0 +1,68 @@
...
+            if cloud_identifier != AbstractStatusProvider.identifier:
+                status_provider_ids.append(cloud_identifier)
+
+        status_provider_ids += \
+            status_config.get_non_cloud_identifiers(config)
+
+        self.status_providers = \
+            self.__get_status_providers(status_provider_ids)
+
+        for k, _v in self.status_providers:  # pylint: disable=invalid-name
```
lorengordon, line 43: Since you aren't using the value...

```suggestion
        for provider in self.status_providers:
```

src/watchmaker/utils/imds/detect/__init__.py
```diff
@@ -0,0 +1,53 @@
...
+    with concurrent.futures.ThreadPoolExecutor(
+            max_workers=MAX_WORKERS) as executor:
+        futures.append(executor.submit(identify, AWSProvider))
+        futures.append(executor.submit(identify, AzureProvider))
+
+    concurrent.futures.wait(futures)
+    for fut in futures:
+        try:
+            result = fut.result()
+            if result != AbstractProvider.identifier:
```
lorengordon, line 32: Under what condition(s) would `AbstractProvider.identifier` be returned, given that we're specifcally calling only for AWSProvider and AzureProvider?

src/watchmaker/utils/imds/detect/providers/azure_provider.py
```diff
@@ -0,0 +1,62 @@
+# -*- coding: utf-8 -*-
+"""AWS Provider."""
```
lorengordon, line 2:
```suggestion
"""Azure Provider."""
```

src/watchmaker/utils/imds/detect/providers/azure_provider.py
```diff
@@ -0,0 +1,62 @@
...
+        self.logger = logger or logging.getLogger(__name__)
+        self.metadata_url = (
+            'http://169.254.169.254/metadata/instance?api-version=2021-02-01'
+        )
+        self.vendor_file = '/sys/class/dmi/id/sys_vendor'
+        self.headers = {'Metadata': 'true'}
+
+    def identify(self):
+        """Identify Azure using all the implemented options."""
+        self.logger.info('Try to identify DO')
```
lorengordon, line 28: What is "DO"?

src/watchmaker/utils/urllib/request_handlers.py
lorengordon: The above change was made, but the goal of that was to allow this module to `import boto3`, without gating it on the try/except.

src/watchmaker/utils/urllib/request_handlers.py
```diff
@@ -46,40 +39,39 @@ def s3_open(self, req):
...
-                'url must be in the format s3://<bucket>/<key>'
-            )
+                "url must be in the format s3://<bucket>/<key>")
 
         try:
             s3_conn = self.s3_conn
         except AttributeError:
             # pylint: disable=attribute-defined-outside-init
-            s3_conn = self.s3_conn = boto3.resource("s3")
+            # pylint: disable=undefined-variable
```
lorengordon, line 48: Presuming boto3 can now be imported, this shouldn't be needed.

src/watchmaker/utils/urllib/request_handlers.py
```diff
@@ -46,40 +39,39 @@ def s3_open(self, req):
...
+            s3_conn = self.s3_conn = boto3.resource("s3")  # type: ignore # noqa F821
```
lorengordon, line 49: Similarly, shouldn't need the noqa here.

tests/test_watchmaker.py
```diff
@@ -8,6 +8,11 @@
 import pytest
 import yaml
 
+try:
+    from unittest.mock import patch
```
lorengordon, line 12: Can you put a comment on why the try/except is necessary? I assume it's a python2/3 thing?

src/watchmaker/utils/imds/detect/__init__.py
```diff
@@ -0,0 +1,53 @@
...
+                results.append(result)
+        except BaseException as ex:
+            exception_list.append(str(ex))
+
+    if len(results) > 1:
+        raise Exception("Detected more than one cloud provider")
```
lorengordon, line 38: Let's raise something from the `watchmaker.exceptions` module, at least WatchmakerError, or more specific.

src/watchmaker/utils/imds/detect/__init__.py
dwc0011: Adding a comment in the code, but a non-cloud environment or unsupported cloud env, for example gcp. The cloud detect module would not detect AWS or AZURE and it would return "unknown".  We are attempting to detect if we are on AWS or AZURE as they are supported, but it does not mean we are on them.

lorengordon: I guess I was just expecting them to return nothing in that case, or an exception. I was thinking of it as something like "assert that I am a valid provider".

lorengordon: Maybe something like:

```
        try:
            results.append(fut.result())
        except InvalidProvider:
            pass
        except BaseException as ex:
            exception_list.append(str(ex))
```

src/watchmaker/utils/imds/detect/providers/azure_provider.py
dwc0011: It was in the code I started from and did not notice, apparently it is Digital Ocean, another cloud provider.

docs/configuration.md
```diff
@@ -208,6 +209,56 @@ Parameters supported by the Yum Worker:
         url: http://someplace.com/my.repo
     ```
 
+### status
+
+Status providers supported for posting watchmaker status.
```
lorengordon, line 214: Suggestion:

Watchmaker supports posting the watchmaker status to status providers. Watchmaker status values are one of: 'Running', 'Failed', or 'Completed'. Each status provider defines what it means to "post the status".. Currently, the supported provider types include: 'aws' and 'azure'. These status providers both post the status as a tag to the instance/VM.

Providers have the ability to detect whether the system is compatible with the provider type. In order to post status, the system running watchmaker must be compatible with the status provider type. For example, the 'azure' provider will be skipped when watchmaker is running on an AWS EC2 instance, and vice versa.

```
```{eval-rst}
.. note::

    Note: Support for the 'azure' status provider is provisional. If you use it and encounter problems, please open an issue on the GitHub repository!
```

lorengordon (approved):
After some testing, there are some enhancements we can make to the behavior. But going ahead and approving, and we can deal with the changes in a separate pr that will be easier to review.

lorengordon:
The threading for the imds check is waiting for all providers, even when one has returned `True`. It seems like it could just cancel pending threads, instead:

```
2022-12-19 22:02:43,978 [watchmaker.utils.imds.detect.providers.aws_provider][INFO ][2120]: Try to identify AWS
2022-12-19 22:02:43,979 [watchmaker.utils.imds.detect.providers.azure_provider][INFO ][2120]: Try to identify Azure
2022-12-19 22:02:43,979 [watchmaker.utils.imds.detect.providers.aws_provider][DEBUG][2120]: Checking AWS metadata
2022-12-19 22:02:43,980 [watchmaker.utils.imds.detect.providers.azure_provider][DEBUG][2120]: Checking Azure vendor file
2022-12-19 22:02:43,982 [watchmaker.utils.imds.detect.providers.azure_provider][DEBUG][2120]: Checking Azure metadata
2022-12-19 22:02:43,987 [backoff][INFO ][2120]: Backing off urlopen_retry(...) for 1.0s (urllib.error.HTTPError: HTTP Error 404: Not Found)
2022-12-19 22:02:44,995 [backoff][INFO ][2120]: Backing off urlopen_retry(...) for 1.4s (urllib.error.HTTPError: HTTP Error 404: Not Found)
2022-12-19 22:02:46,385 [backoff][INFO ][2120]: Backing off urlopen_retry(...) for 3.3s (urllib.error.HTTPError: HTTP Error 404: Not Found)
2022-12-19 22:02:49,708 [backoff][INFO ][2120]: Backing off urlopen_retry(...) for 0.2s (urllib.error.HTTPError: HTTP Error 404: Not Found)
2022-12-19 22:02:49,963 [backoff][ERROR][2120]: Giving up urlopen_retry(...) after 5 tries (urllib.error.HTTPError: HTTP Error 404: Not Found)
2022-12-19 22:02:49,963 [watchmaker.utils.imds.detect.providers.azure_provider][ERROR][2120]: HTTP Error 404: Not Found
2022-12-19 22:02:49,995 [watchmaker.status.providers.aws][DEBUG][2120]: Tagging AWS Resource if HAS_BOTO3
```

Also, the behavior of the imds provider is slightly different on Windows vs Linux. Windows _always_ hits the `urlopen_retry()` logic in the azure imds provider, and retries 5 times. Linux _never_ hits the `urlopen_retry()` logic, even though it probably should be?

PR #2358, opened 2022-10-28
Reverts pyinstaller version and updates mergify config to handle future pyinstaller dependabot updates

eemperor:


eemperor:
go codebuild go

lorengordon:
@eemperor Looks like the last _working_ version of pyinstaller might have been 5.3.0?

https://github.com/plus3it/watchmaker/pulls?q=is%3Apr+sort%3Aupdated-desc+is%3Aclosed+pyinstaller

eemperor:
> @eemperor Looks like the last _working_ version of pyinstaller might have been 5.3?
> 
> https://github.com/plus3it/watchmaker/pulls?q=is%3Apr+sort%3Aupdated-desc+is%3Aclosed+pyinstaller

eemperor:
Oops.  Clicked the wrong option...

eemperor:
go codebuild go

eemperor:
go codebuild go

.mergify.yml
```diff
@@ -4,23 +4,35 @@ pull_request_rules:
       - author~=dependabot\[bot\]|dependabot-preview\[bot\]
       - status-success=Travis CI - Pull Request
       - status-success=continuous-integration.azure-devops
+      - or:
+          - -title~=(?i)pyinstaller
```
lorengordon, line 8: The `-` is so hard to see. Can we make the condition more obvious using the `not:` condition?

lorengordon: Also, do we need to match `(?i).*pyinstaller.*`?

.mergify.yml
```diff
@@ -4,23 +4,35 @@ pull_request_rules:
...
         type: APPROVE
-  # For salt-formula updates, trigger codebuild, merge on codebuild status
+
+  # For salt-formula or pyinstaller updates, trigger codebuild, merge on codebuild status
   - name: write issue comment to start codebuild job
     conditions:
       - author~=dependabot\[bot\]|dependabot-preview\[bot\]
-      - label=submodules
-      - "#approved-reviews-by>=1"
+      - or:
```
lorengordon, line 21: i think we can flip this, so both cases require the approval rule first:

```suggestion
      - and:
        - "#approved-reviews-by>=1"
        - or:
          - label=submodules
          - title~=(?i)pyinstaller
```

.mergify.yml
```diff
@@ -34,6 +46,7 @@ pull_request_rules:
       - author~=dependabot\[bot\]|dependabot-preview\[bot\]
       - "#approved-reviews-by>=1"
       - label!=submodules
+      - -title~=(?i)pyinstaller
```
lorengordon, line 49: perhaps use `not:` here also?

eemperor: The mergify check doesn't seem to like using `not:` under another condition?

lorengordon: That is really weird. I guess they don't specifically show how this is supposed to work, but it does say the depth is limited to 8...?

https://docs.mergify.com/conditions/#combining-conditions-with-operators

eemperor: Yeah.  It's strange.  Their example only has `not:` at the top.

.mergify.yml
```diff
@@ -4,23 +4,35 @@ pull_request_rules:
       - author~=dependabot\[bot\]|dependabot-preview\[bot\]
       - status-success=Travis CI - Pull Request
       - status-success=continuous-integration.azure-devops
+      - or:
```
lorengordon, line 7: I don't think we need this block after all. Without it, the CI will run, the standalone will be created, mergify will "approve", then the next rule will comment, and the merge rule will not merge unless the codebuild jobs post success.

eemperor: Okay.  I think you're right.  Otherwise, there would be no approve rule for the submodules or pyinstaller.

eemperor: Fixed it.

lorengordon (approved):
Let's give it a try!

Issue #72, opened 2016-09-01
Disable salt-minion service on Linux hosts


ferricoxide:
By default, Watchmaker's installation of the salt packages installs the `salt-minion` service. This service should be stopped and disabled:
- EL6: `chkconfig salt-minion off ; service salt-minion stop`
- EL7: `systemctl stop salt-minion ; systemctl disable salt-minion ; systemctl mask salt-minion`

lorengordon:
@ferricoxide can you test if we can use salt itself to properly stop and disable the salt-minion on EL6 and EL7, rather than the platform-specific calls? That way we don't need logic to handle different platforms...

`salt-call --local service.stop salt-minion && salt-call --local service.disable salt-minion`

ferricoxide:
Just for clarity: is the above testing with an eye towards adding a standardized salt-state to the deployed state sets or something to be run as a stand-alone part of UserData? If the latter, is there a specific value to using this method versus generically-prescribing "shut off the salt-minion service at launch time"?

lorengordon:
We would incorporate that into watchmaker, just after the install step and (probably) before running the highstate. We already have logic to run salt commands, such as setting grains and applying states, so this is just extending that logic to run one more salt command.

ferricoxide:
Cool. I'll try to give it a quick-check, this morning.

KevinPlus3:
Any success on these salt commands ... I am just reading this and realizing I could wrap this into the current PR.

lorengordon:
Ahh, no, that PR is large enough. Prefer keeping PRs small and focused on a single issue so they are easier to review.

lorengordon:
Thinking about this more, we should test prior to salt install if salt is already present/enabled/running and preserve that state. Starting to see people using salt for their own purposes, and would prefer to avoid stepping on their toes...

PR #2389, opened 2022-11-15
Add Support for CentOS 8 Stream

### LINKED ISSUE #2380: [Enhancement] Add Support for CentOS 8 Stream
Currently, watchmaker is failing in the [`yum.py`](https://github.com/plus3it/watchmaker/blob/main/src/watchmaker/workers/yum.py) module's [`get_dist_info` function](https://github.com/plus3it/watchmaker/blob/main/src/watchmaker/workers/yum.py#L55-L105). Presumably, the format of CentOS 8 stream's `/etc/system-release` file doesn't adhere to parsing expectations.

ferricoxide:
Closes #2380 

- Validated against CentOS 8 Stream spel AMI `ami-0a6c4db379b52a51a`
- Validated against RHEL 8 spel (beta) AMI `ami-02e6c5efdce6956d`

**Note:** Due the changes in methodology used to identify target EC2s' distro-name, it will be necessary to update any free-standing `config.yaml` files to include `rhel` in the current `redhat` and `centos` config-map (see 480257e in this PR). The `redhat` map-member is preserved for backwards-compatibility

### DIFF
--- a/src/watchmaker/static/config.yaml
+++ b/src/watchmaker/static/config.yaml
@@ -25,7 +25,6 @@ linux:
         #SaltEL6:
         - dist:
             - redhat
-            - rhel
             - centos
           el_version: 6
           url: https://watchmaker.cloudarmor.io/yum.defs/saltstack/salt/2019.2.8/salt-reposync-el6.repo
@@ -35,14 +34,12 @@ linux:
         #SaltEL7:
         - dist:
             - redhat
-            - rhel
             - centos
           el_version: 7
           url: https://watchmaker.cloudarmor.io/yum.defs/saltstack/salt/3004.2/salt-reposync-el7-python3.repo
         #SaltEL8:
         - dist:
             - redhat
-            - rhel
             - centos
           el_version: 8
           url: https://watchmaker.cloudarmor.io/yum.defs/saltstack/salt/3004.2/salt-reposync-el8-python3.repo
--- a/src/watchmaker/static/config.yaml
+++ b/src/watchmaker/static/config.yaml
@@ -25,6 +25,7 @@ linux:
         #SaltEL6:
         - dist:
             - redhat
+            - rhel
             - centos
           el_version: 6
           url: https://watchmaker.cloudarmor.io/yum.defs/saltstack/salt/2019.2.8/salt-reposync-el6.repo
@@ -34,12 +35,14 @@ linux:
         #SaltEL7:
         - dist:
             - redhat
+            - rhel
             - centos
           el_version: 7
           url: https://watchmaker.cloudarmor.io/yum.defs/saltstack/salt/3004.2/salt-reposync-el7-python3.repo
         #SaltEL8:
         - dist:
             - redhat
+            - rhel
             - centos
           el_version: 8
           url: https://watchmaker.cloudarmor.io/yum.defs/saltstack/salt/3004.2/salt-reposync-el8-python3.repo
--- a/src/watchmaker/workers/yum.py
+++ b/src/watchmaker/workers/yum.py
@@ -3,7 +3,6 @@
 from __future__ import (absolute_import, division, print_function,
                         unicode_literals, with_statement)
 
-import io
 import re
 
 import distro
--- a/src/watchmaker/workers/yum.py
+++ b/src/watchmaker/workers/yum.py
@@ -55,12 +55,12 @@ class Yum(WorkerBase, LinuxPlatformManager):
 
     def get_dist_info(self):
         """Validate the Linux distro and return info about the distribution."""
-        os_dist = dist.id()
+        dist = dist.id()
         version = dist.version()[0]
         el_version = None
 
         # Determine el_version
-        if os_dist == 'amazon':
+        if dist == 'amazon':
             el_version = self._get_amazon_el_version(version)
         else:
             el_version = dist.version()[0]
@@ -74,7 +74,7 @@ class Yum(WorkerBase, LinuxPlatformManager):
             raise WatchmakerError(msg)
 
         dist_info = {
-            'dist': os_dist,
+            'dist': dist,
             'el_version': el_version
         }
         self.log.debug('dist_info=%s', dist_info)
--- a/src/watchmaker/workers/yum.py
+++ b/src/watchmaker/workers/yum.py
@@ -33,7 +33,7 @@ class Yum(WorkerBase, LinuxPlatformManager):
     DIST_PATTERN = re.compile(
         r"^({0})"
         "(?:[^0-9]+)"
-        r"([\d]+[.][\d]+)"
+        r"([\d]|[\d]+[.][\d])"
         "(?:.*)"
         .format('|'.join(SUPPORTED_DISTS))
     )
--- a/src/watchmaker/workers/yum.py
+++ b/src/watchmaker/workers/yum.py
@@ -6,6 +6,7 @@ from __future__ import (absolute_import, division, print_function,
 import io
 import re
 
+import distro
 import six
 
 import watchmaker.utils
@@ -25,7 +26,7 @@ class Yum(WorkerBase, LinuxPlatformManager):
 
     """
 
-    SUPPORTED_DISTS = ('amazon', 'centos', 'red hat')
+    SUPPORTED_DISTS = ('amazon', 'centos', 'red hat', 'rhel')
 
     # Pattern used to match against the first line of /etc/system-release. A
     # match will contain two groups: the dist name (e.g. 'red hat' or 'amazon')
@@ -54,40 +55,15 @@ class Yum(WorkerBase, LinuxPlatformManager):
 
     def get_dist_info(self):
         """Validate the Linux distro and return info about the distribution."""
-        dist = None
-        version = None
+        dist = dist.id()
+        version = dist.version()[0]
         el_version = None
-
-        # Read first line from /etc/system-release
-        try:
-            with io.open('/etc/system-release', encoding='utf8') as fh_:
-                release = fh_.readline().strip()
-        except Exception:
-            self.log.critical(
-                'Failed to read /etc/system-release. Cannot determine system '
-                'distribution!'
-            )
-            raise
-
-        # Search the release file for a match against _supported_dists
-        matched = self.DIST_PATTERN.search(release.lower())
-        if matched is None:
-            # Release not supported, exit with error
-            msg = (
-                'Unsupported OS distribution. OS must be one of: {0}'
-                .format(', '.join(self.SUPPORTED_DISTS))
-            )
-            self.log.critical(msg)
-            raise WatchmakerError(msg)
-
-        # Assign dist,version from the match groups tuple, removing any spaces
-        dist, version = (x.replace(' ', '') for x in matched.groups())
 
         # Determine el_version
         if dist == 'amazon':
             el_version = self._get_amazon_el_version(version)
         else:
-            el_version = version.split('.')[0]
+            el_version = dist.version()[0]
 
         if el_version is None:
             msg = (
--- a/src/watchmaker/workers/yum.py
+++ b/src/watchmaker/workers/yum.py
@@ -55,12 +55,12 @@ class Yum(WorkerBase, LinuxPlatformManager):
 
     def get_dist_info(self):
         """Validate the Linux distro and return info about the distribution."""
-        dist = dist.id()
+        os_dist = dist.id()
         version = dist.version()[0]
         el_version = None
 
         # Determine el_version
-        if dist == 'amazon':
+        if os_dist == 'amazon':
             el_version = self._get_amazon_el_version(version)
         else:
             el_version = dist.version()[0]
@@ -74,7 +74,7 @@ class Yum(WorkerBase, LinuxPlatformManager):
             raise WatchmakerError(msg)
 
         dist_info = {
-            'dist': dist,
+            'dist': os_dist,
             'el_version': el_version
         }
         self.log.debug('dist_info=%s', dist_info)
--- a/src/watchmaker/workers/yum.py
+++ b/src/watchmaker/workers/yum.py
@@ -55,15 +55,15 @@ class Yum(WorkerBase, LinuxPlatformManager):
 
     def get_dist_info(self):
         """Validate the Linux distro and return info about the distribution."""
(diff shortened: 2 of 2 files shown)


ferricoxide:
go codebuild go

ferricoxide:
go codebuild go

src/watchmaker/static/config.yaml
```diff
@@ -34,12 +35,14 @@ linux:
...
         - dist:
             - redhat
+            - rhel
             - centos
           el_version: 7
           url: https://watchmaker.cloudarmor.io/yum.defs/saltstack/salt/3004.2/salt-reposync-el7-python3.repo
         #SaltEL8:
         - dist:
             - redhat
+            - rhel
```
lorengordon, line 45: Hmm, this doesn't quite map out correctly... this config was matching on `redhat`, but `SUPPORTED_DISTS` was matching `red hat` (with a space). That implies to me that we're missing something about how the config is mapped to the dist. But I haven't been in this area of the code in a long time, so I can't say exactly what we're missing without tracing things more thoroughly.

src/watchmaker/workers/yum.py
```diff
@@ -25,15 +25,15 @@ class Yum(WorkerBase, LinuxPlatformManager):
...
+    SUPPORTED_DISTS = ('amazon', 'centos', 'red hat', 'rhel')
 
     # Pattern used to match against the first line of /etc/system-release. A
     # match will contain two groups: the dist name (e.g. 'red hat' or 'amazon')
     # and the dist version (e.g. '6.8' or '2016.09').
     DIST_PATTERN = re.compile(
         r"^({0})"
         "(?:[^0-9]+)"
-        r"([\d]+[.][\d]+)"
+        r"([\d]|[\d]+[.][\d])"
```
lorengordon, line 36: With the other changes, `DIST_PATTERN` is not used at all, so can be removed entirely.

src/watchmaker/workers/yum.py
```diff
@@ -54,40 +54,15 @@ def _get_amazon_el_version(self, version):
...
-        # Search the release file for a match against _supported_dists
-        matched = self.DIST_PATTERN.search(release.lower())
-        if matched is None:
-            # Release not supported, exit with error
-            msg = (
-                'Unsupported OS distribution. OS must be one of: {0}'
-                .format(', '.join(self.SUPPORTED_DISTS))
-            )
-            self.log.critical(msg)
-            raise WatchmakerError(msg)
```
lorengordon, line 81: We should still raise and exit with error if `dist` is not one of the `SUPPORTED_DISTS`.

src/watchmaker/workers/yum.py
ferricoxide: Was _thinking_ that, but "wasn't eating anything". 

Removed.

src/watchmaker/workers/yum.py
ferricoxide: Clapper's whacking my test-EC2s. This _should_ satisfy the requirement:

~~~
       # Error if 'dist' does not found in SUPPORTED_DISTS
       if dist not in SUPPORTED_DISTS
           # Release not supported, exit with error
           msg = (
               'Unsupported OS distribution. OS must be one of: {0}'
               .format(', '.join(self.SUPPORTED_DISTS))
           )
           self.log.critical(msg)
           raise WatchmakerError(msg)
~~~

?

At any rate, once it's after 1830, I'll do test-launches to validate.

ferricoxide:
go codebuild go

src/watchmaker/workers/yum.py
```diff
@@ -54,40 +41,15 @@ def _get_amazon_el_version(self, version):
...
-
-        # Assign dist,version from the match groups tuple, removing any spaces
-        dist, version = (x.replace(' ', '') for x in matched.groups())
```
lorengordon, line 84: Ahh, lol, _this_ is what was changing `red hat` to `redhat`. Ok, now that we're using `distro`, I think we can just re-map `rhel` to `redhat` explicitly?

src/watchmaker/workers/yum.py
lorengordon: That looks right, yes

ferricoxide:
go codebuild go

src/watchmaker/workers/yum.py
ferricoxide: Err.. is that outdated code-content? Doing a search in my editor for `x.replace` gives me:
```
E486: Pattern not found: x.replace
```

src/watchmaker/workers/yum.py
ferricoxide: Alright, bout to launch new EC2s to see if this makes anything vomit in either RHEL 8 or CentOS 8-stream.

ferricoxide: Ok, so `SUPPORTED_DISTS` is defined in the `Yum` class but it's not available to the `get_dist_info` function in that class?

ferricoxide: Oh... `self.SUPPORTED_DISTS`?

src/watchmaker/workers/yum.py
lorengordon: That makes sense, since the comment is on a removed line. ;)

It was intended as providing further details and context on this comment, https://github.com/plus3it/watchmaker/pull/2389/files#r1023345795

src/watchmaker/workers/yum.py
ferricoxide: Alright, working nao.

ferricoxide:
go codebuild go

src/watchmaker/workers/yum.py
ferricoxide: Ok, so… Resolve?

src/watchmaker/static/config.yaml
ferricoxide: /shrug

ferricoxide: Within the code, itself, the explicit dist-references are pretty limited:

~~~
$ find . -type f -print0 | xargs -0 grep -E "red(| )hat"
./src/watchmaker/workers/yum.py:    SUPPORTED_DISTS = ('amazon', 'centos', 'red hat', 'rhel')
./src/watchmaker/static/config.yaml:            - redhat
./src/watchmaker/static/config.yaml:            - redhat
./src/watchmaker/static/config.yaml:            - redhat
./docs/configuration.md:        redhat, centos, and amazon.)
./docs/configuration.md:          - redhat
./docs/configuration.md:            - redhat
./docs/configuration.md:            - redhat
~~~

So, wouldn't really know where to look.

src/watchmaker/workers/yum.py
lorengordon: I think we'll be better off creating a map of `distro.id()` values to watchmaker "config" values. That would mean something like changing `SUPPORTED_DISTS` to be a map:

```
    SUPPORTED_DISTS = {
        'amazon': 'amazon',
        'centos': 'centos',
        'rhel': 'redhat'
    }
```

and then we lookup the dist from that map:

```
        try:
            dist = self.SUPPORTED_DISTS[distro.id()]
        except KeyError:
            # Release not supported, exit with error
            msg = (
                'Unsupported OS distribution. OS must be one of: {0}'
                .format(', '.join(self.SUPPORTED_DISTS.keys()))
            )
            self.log.critical(msg)
            raise WatchmakerError(msg)
```

ferricoxide: 👍 

Cool. I'll give that a shot – possibly as early as later this afternoon or early this evening.

src/watchmaker/static/config.yaml
ferricoxide: Ah. When I'd clicked on the link, the page that rendered was hard to follow. May have been a rendering-blip. 🤷

src/watchmaker/workers/yum.py
ferricoxide: Ok, so, we're initially defining `dist` as the string returned by `distro.id()` (at line #72), but subsequently redefining it with the `try`/`except` block. Is that what we actually want to do? Or do we want to replace the current:

~~~
    def get_dist_info(self):
        """Validate the Linux distro and return info about the distribution."""
        dist = distro.id()
        version = distro.version()[0]
        el_version = None

        # Determine el_version
        if dist == 'amazon':
            el_version = self._get_amazon_el_version(version)
        else:
            el_version = distro.version()[0]

        if el_version is None:
            msg = (
                'Unsupported OS version! dist = {0}, version = {1}.'
                .format(dist, version)
            )
            self.log.critical(msg)
            raise WatchmakerError(msg)

        # Error if 'dist' is not found in SUPPORTED_DISTS
        if dist not in self.SUPPORTED_DISTS:
            # Release not supported, exit with error
            msg = (
                'Unsupported OS distribution. OS must be one of: {0}'
                .format(', '.join(self.SUPPORTED_DISTS))
            )
            self.log.critical(msg)
            raise WatchmakerError(msg)

        dist_info = {
            'dist': dist,
            'el_version': el_version
        }
        self.log.debug('dist_info=%s', dist_info)
        return dist_info
~~~

With something more like:

~~~
    def get_dist_info(self):
        """Validate the Linux distro and return info about the distribution."""
        # Error if 'dist' is not found in SUPPORTED_DISTS
        try:
            dist = self.SUPPORTED_DISTS[distro.id()]
        except KeyError:
            # Release not supported, exit with error
            msg = (
                'Unsupported OS distribution. OS must be one of: {0}'
                .format(', '.join(self.SUPPORTED_DISTS.keys()))
            )
            self.log.critical(msg)
            raise WatchmakerError(msg)
        version = distro.version()[0]
        el_version = None

        # Determine el_version
        if dist == 'amazon':
            el_version = self._get_amazon_el_version(version)
        else:
            el_version = distro.version()[0]

        if el_version is None:
            msg = (
                'Unsupported OS version! dist = {0}, version = {1}.'
                .format(dist, version)
            )
            self.log.critical(msg)
            raise WatchmakerError(msg)

        dist_info = {
            'dist': dist,
            'el_version': el_version
        }
        self.log.debug('dist_info=%s', dist_info)
        return dist_info
~~~

I'm not generally a fan of overloading variables, so, I'll implement the latter and see what breaks.

ferricoxide: Alright. The latter works.

ferricoxide:
go codebuild go

ferricoxide:
go codebuild go

src/watchmaker/static/config.yaml
ferricoxide: Alright, ea9cf8c appears to address this (tested with a `config.yaml` that omits the `rhel` key)

ferricoxide: Oh... need to undo this change. 😛

ferricoxide:
go codebuild go

src/watchmaker/workers/yum.py
lorengordon: Indeed, my suggestion was targeted at replacing the initial assignment of `dist = distro.id()`.

src/watchmaker/workers/yum.py
```diff
@@ -54,40 +45,25 @@ def _get_amazon_el_version(self, version):
 
     def get_dist_info(self):
         """Validate the Linux distro and return info about the distribution."""
-        dist = None
```
lorengordon, line 57: These three assignments for `dist`, `version`, and `el_version` should probably be kept here at the top, to ensure they are set and have values (`None`). Otherwise, they are set in `try` or `if` blocks, which means they may not be set at all if there is an error or an unexpected condition.

src/watchmaker/workers/yum.py
```diff
@@ -54,40 +45,25 @@ def _get_amazon_el_version(self, version):
...
-                .format(', '.join(self.SUPPORTED_DISTS))
+                .format(', '.join(self.SUPPORTED_DISTS.keys()))
             )
             self.log.critical(msg)
             raise WatchmakerError(msg)
-
-        # Assign dist,version from the match groups tuple, removing any spaces
-        dist, version = (x.replace(' ', '') for x in matched.groups())
+        version = distro.version()[0]
+        el_version = None
```
lorengordon, line 60: Once this assignment is again handled at the top of the function, this line can be removed.

src/watchmaker/workers/yum.py
ferricoxide: The `except` should abort the overall execution, though. (and the `version` and `el_version` are still kept at the top, just that `dist` is now a multi-line value-setting.

src/watchmaker/workers/yum.py
ferricoxide: "This line"?

lorengordon: Comments code code in github pull requests are on lines. So yes, "this line" is the line this comment is attached to. It's difficult to refer to specific line numbers, since as changes are made, the line number will change. But at the moment, it is line 60.

src/watchmaker/workers/yum.py
```diff
@@ -104,6 +70,20 @@ def get_dist_info(self):
         self.log.debug('dist_info=%s', dist_info)
         return dist_info
 
+    def get_mapped_dist_name(self):
+        """Return a normalized dist-name value."""
+        # Error if 'dist' is not found in SUPPORTED_DISTS
+        try:
+            dist = self.SUPPORTED_DISTS[distro.id()]
```
lorengordon, line 77:
```suggestion
            return self.SUPPORTED_DISTS[distro.id()]
```

ferricoxide: DAMMIT. 😛

src/watchmaker/workers/yum.py
lorengordon: The change to use a function simplifies it enough for me. Just need to fix it so it returns the value...

src/watchmaker/workers/yum.py
lorengordon: Become moot, since the refactor to use a function.

ferricoxide:
go codebuild go

lorengordon:
go codebuild go

lorengordon:
go codebuild go

lorengordon (approved):
🚀

Issue #2461, opened 2022-12-20
Provider checks should pass all prereqs before attempting detection


dwc0011:
Providers often have certain prereqs on an environment that must be met in order to work.  Such as boto3 for AWS providers or azure python libs for Azure providers.  Before watchmaker attempts to detect if we are on a specific provider we can determine if the the prereqs are met and if not, bypass the detection.

dwc0011:
Beginning work on this @ [dwc0011-provider-prereq-checks](https://github.com/dwc0011/watchmaker/tree/dwc0011-provider-prereq-checks)

PR #2468, opened 2022-12-21
Adds supported providers and their prereqs status map

dwc0011:
This PR fixes [#2461](https://github.com/plus3it/watchmaker/issues/2461).

It adds a list of providers and the prereq status, this prereqs is based on if the libraries required are present.


dwc0011:
Ran tests against Linux with and without boto3 and azure python libs installed

docs/configuration.md
```diff
@@ -76,6 +76,90 @@ systems.
 Section for Worker configurations that should only be applied to Windows-based
 systems.
 
+### status
+
+Watchmaker supports posting the watchmaker status to status providers. Watchmaker status values are one of: 'Running', 'Failed', or 'Completed'. Each status provider defines what it means to "post the status".. Currently, the supported provider types include: 'aws' and 'azure'. These status providers both post the status as a tag to the instance/VM.
+
+Providers have the ability to detect whether the system is compatible with the provider type. In order to post status, the system running watchmaker must be compatible with the status provider type. For example, the 'azure' provider will be skipped when watchmaker is running on an AWS EC2 instance, and vice versa.
+
+Prerequisites for AWS:
```
lorengordon, line 85: This doc section was reworked in https://github.com/plus3it/watchmaker/pull/2585. I'd suggest skimming that changeset for the diff and updating this to match.

src/watchmaker/status/__init__.py
```diff
@@ -37,11 +38,34 @@ def initialize(self, config=None):
...
+            # Check for supported providers missing prereqs
+            missing_prereq_cloud_ids = (
+                status_config.get_req_cloud_ids_wo_prereqs(config)
+            )
+
+            if missing_prereq_cloud_ids:
+                # Detect provider
+                cloud_identifier = provider(missing_prereq_cloud_ids)
+
+                if cloud_identifier != AbstractStatusProvider.identifier:
```
lorengordon, line 61: I kinda hate the deeply nested `if` statements. Any easy ideas on something a little flatter or more declarative?

src/watchmaker/utils/imds/detect/providers/azure_provider.py
```diff
@@ -47,18 +47,24 @@ def check_vendor_file(self):
...
+            self.logger.debug("Vendor file does not exist")
+            return False
+
+        self.logger.debug("Vendor file exists")
+        vendor_file_path = utils.uri_from_filepath(self.vendor_file)
+        try:
+            check_vendor_file = utils.urlopen_retry(
+                vendor_file_path, self.DEFAULT_TIMEOUT
+            )
+            if b"Microsoft Corporation" in check_vendor_file.read():
```
lorengordon, line 63: Do you happen to remember why this needs to be `b""` now?

dwc0011: The read method returns bytes, I just updated the aws one to convert the strnig to bytes, tested it, and it worked.  The aws code path always passed on metadata server and never checked the file, had to force the test.  When we check for azure the detect code checked the file before the metadata server and since that file exists on linux, regardless of AWS or Azure, we did hit the bytes like error not str, so I added the bytes conversion.

src/watchmaker/status/__init__.py
dwc0011: Check the latest and see if you like better.

docs/configuration.md
```diff
@@ -76,6 +76,65 @@ systems.
 Section for Worker configurations that should only be applied to Windows-based
 systems.
 
+### status
+
```
lorengordon, line 80: Missing the "intro" paragraph from before, trying to describe what the "status" actually does...

```
Watchmaker supports posting the watchmaker status to status providers. Watchmaker
status values are one of: 'Running', 'Failed', or 'Completed'. Each status provider
defines what it means to "post the status". Currently, the supported provider
types include: 'aws' and 'azure'. These status providers both post the status as
a tag to the instance/VM.
```

docs/configuration.md
```diff
@@ -209,120 +268,6 @@ Parameters supported by the Yum Worker:
...
-                "StringLike": {
-                    "ec2:SourceInstanceARN": "${ec2:SourceInstanceARN}"
-                }
-            }
-        }
-    ]
-}
-```
-
-## Downloading config files from Amazon S3
```
lorengordon, line 302: This section shouldn't be deleted

docs/configuration.md
dwc0011: Fixed

src/watchmaker/status/__init__.py
lorengordon: 🎆

dwc0011:
Tested all the various permutations of config (required, not required, etc) on linux AWS and worked as expected.

lorengordon (approved):
🚢

PR #2536, opened 2023-01-25
Removes gravitybee in standalone build, uses pyinstaller directly

lorengordon:
Currently the standalone does not run on EL8 with FIPS enabled. It will just fail with this message, even just getting the version:

```
# ./watchmaker --version
fips.c(145): OpenSSL internal error, assertion failed: FATAL FIPS SELFTEST FAILURE
Aborted (core dumped)
```

This pr is removing gravitybee, because the abstraction was making it too difficult to understand exactly how the standalone was being built and modify the contents of the package. Pyinstaller is being used directly now, so we can adjust the package by modifying `ci/pyinstaller/hook-watchmaker.py`, or by modifying the pyinstaller command in `ci/build.sh`.


lorengordon:
go codebuild go

lorengordon:
Still need to spend some time with the Windows package. That is still using gravitybee, at the moment.

lorengordon:
go codebuild go

ci/build.sh
```diff
@@ -3,8 +3,17 @@ set -eu -o pipefail
 
 export VIRTUAL_ENV_DIR=venv
 
+VERSION=$(grep "version =" setup.cfg | sed 's/^.*= //')
```
ferricoxide, line 6: Can save yourself a `grep` with a denser `sed` invocation:

```suggestion
VERSION=$( sed -n 's/^\s*version\s*=\s*\(.*\)/\1/p' setup.cfg )
```

lorengordon:
go codebuild go

ci/build.sh
lorengordon: I'm going to leave the grep/sed combo, since we're using that most everywhere, and the consistency is worth more than doing it all in a single tool.

eemperor (approved):
Okay.  Looks good. :shipit:

PR #2561, opened 2023-02-04
Readds data dirs so pyinstaller picks up non-package .py files

lorengordon:


ci/pyinstaller/hook-watchmaker.py
```diff
@@ -6,7 +6,13 @@
 from PyInstaller.utils.hooks import (collect_data_files, collect_dynamic_libs,
                                      collect_submodules, copy_metadata)
 
-datas = []
+datas = [
+    ("src/watchmaker/static", "watchmaker/static"),
```
eemperor, line 10: Minor nit.  Can you match the formatting for dirs?

lorengordon: This is how `black` formats the file, I don't think we should fight with the tool.

lorengordon:
go codebuild go

ci/pyinstaller/hook-watchmaker.py
eemperor: That's interesting.  Okay, it's good with me.

eemperor (approved):

PR #2617, opened 2023-02-22
Isort black profile

dwc0011:
configure isort to use the black profile to prevent conflicting formatting


lorengordon:
go codebuild go

lorengordon (approved):
🚀

PR #2684, opened 2023-03-14
Version check

ihenyene:
Added method to check salt-call version and skip install if version matches.

### DIFF
--- a/src/watchmaker/workers/salt.py
+++ b/src/watchmaker/workers/salt.py
@@ -720,6 +720,7 @@ class SaltLinux(SaltBase, LinuxPlatformManager):
         self.bootstrap_source = \
             kwargs.pop('bootstrap_source', None) or ''
         self.git_repo = kwargs.pop('git_repo', None) or ''
+        self.salt_content_path = kwargs.pop('salt_content_path', None) or ''
         self.salt_version = kwargs.pop('salt_version', None) or ''
 
         # Enforce lowercase and replace spaces with ^ in Linux
@@ -774,10 +775,22 @@ class SaltLinux(SaltBase, LinuxPlatformManager):
                     'parameter `git_repo` was not provided.'
                 )
 
+    def _check_salt_version(self):
+        current_salt_version = self.call_process(['salt-call', '--version'])
+        if self.salt_version and self.salt_version in current_salt_version:
+            self.log.info('Salt is already installed.')
+            return True
+        self.log.info('salt version is not installed')
+        return False
+
     def _install_package(self):
         if self.install_method.lower() == 'yum':
             self._install_from_yum(self.yum_pkgs)
         elif self.install_method.lower() == 'git':
+            if self._check_salt_version():
+                return
+
+            self.log.info('Starting Salt installation')
             salt_bootstrap_filename = os.sep.join((
                 self.working_dir,
                 watchmaker.utils.basename_from_uri(self.bootstrap_source)
@@ -982,10 +995,14 @@ class SaltWindows(SaltBase, WindowsPlatformManager):
         salt_svc = 'salt-minion'
         if os.path.exists(self.salt_call):
             salt_running, salt_enabled = self.service_status(salt_svc)
-        self._install_package()
-        if self.pip_install:
-            self._install_pip_packages()
-        self.salt_call = SaltWindows._get_salt_call()
+            self.log.info('Skipping installation, already installed.')
+        else:
+            self.log.info('Installing Salt minion...')
+            self._install_package()
+            if self.pip_install:
+                self._install_pip_packages()
+            self.salt_call = SaltWindows._get_salt_call()
+            self.log.info("Salt minion installation completed successfully.")
         salt_stopped = self.service_stop(salt_svc)
         self._build_salt_formula(self.salt_srv)
         if salt_enabled:


src/watchmaker/workers/salt.py
```diff
@@ -774,10 +775,22 @@ def _configuration_validation(self):
...
+        return False
+
     def _install_package(self):
         if self.install_method.lower() == 'yum':
             self._install_from_yum(self.yum_pkgs)
         elif self.install_method.lower() == 'git':
+            if self._check_salt_version():
+                return
+
+            self.log.info('Starting Salt installation')
```
lorengordon, line 793: We can move the check higher, so it applies to all install methods, not just `git`.

```suggestion
    def _install_package(self):
        if os.path.exists(self.salt_call) and self._check_salt_version():
            return

        if self.install_method.lower() == 'yum':
            self._install_from_yum(self.yum_pkgs)
        elif self.install_method.lower() == 'git':
```

src/watchmaker/workers/salt.py
```diff
@@ -720,6 +720,7 @@ def __init__(self, *args, **kwargs):
         self.bootstrap_source = \
             kwargs.pop('bootstrap_source', None) or ''
         self.git_repo = kwargs.pop('git_repo', None) or ''
+        self.salt_content_path = kwargs.pop('salt_content_path', None) or ''
```
lorengordon, line 723: This seems unrelated to the rest of the changes. It wasn't needed before, so I don't think it is needed?

```suggestion
```

lorengordon (changes requested):

src/watchmaker/workers/salt.py
```diff
@@ -774,10 +774,22 @@ def _configuration_validation(self):
                     'parameter `git_repo` was not provided.'
                 )
 
+    def _check_salt_version(self):
+        current_salt_version = self.call_process(['salt-call', '--version'])
+        if self.salt_version and self.salt_version in current_salt_version:
+            self.log.info('Salt is already installed.')
```
lorengordon, line 780:
```suggestion
            self.log.info('Salt version %s is already installed.', self.salt_version)
```

src/watchmaker/workers/salt.py
```diff
@@ -774,10 +774,22 @@ def _configuration_validation(self):
...
+            return True
+        self.log.info('salt version is not installed')
```
lorengordon, line 782:
```suggestion
        self.log.info('Salt version was not specified or does not match.')
```

src/watchmaker/workers/salt.py
```diff
@@ -774,10 +774,22 @@ def _configuration_validation(self):
...
+        return False
+
     def _install_package(self):
+        if os.path.exists(self.salt_call) and self._check_salt_version():
+            return
         if self.install_method.lower() == 'yum':
             self._install_from_yum(self.yum_pkgs)
         elif self.install_method.lower() == 'git':
+
+            self.log.info('Starting Salt installation')
```
lorengordon, line 792: If you move this up, just above line 788, then it will log this for every salt install method and not just `git`.

src/watchmaker/workers/salt.py
```diff
@@ -774,10 +774,22 @@ def _configuration_validation(self):
                     'parameter `git_repo` was not provided.'
                 )
 
+    def _check_salt_version(self):
```
lorengordon, line 777: This method is currently in `SaltLinux`, which means it is only available to Linux systems. We need this check on Windows also. So let's move this method to `SaltBase`.

src/watchmaker/workers/salt.py
```diff
@@ -986,6 +998,7 @@ def install(self):
         if self.pip_install:
             self._install_pip_packages()
         self.salt_call = SaltWindows._get_salt_call()
+        self.log.info("Salt minion installation completed successfully.")
```
lorengordon, line 1001: I think we can remove this log line, since we're adding a check for the salt version, and might not have installed salt at all. That means this line would be misleading.

```suggestion
```

lorengordon (changes requested):
Currently, the method `_check_salt_version()` is only in the `SaltLinux` class. We want to perform this check on both Windows and Linux, so the method needs to move to `SaltBase`.

Also, there is an `_install_package()` method in _both_ `SaltLinux` and `SaltWindows`. Both methods need to be updated to test the salt version and return early if the version matches.

src/watchmaker/workers/salt.py
```diff
@@ -240,6 +241,19 @@ def _prepare_for_install(self):
         ) as fh_:
             yaml.safe_dump(self.salt_conf, fh_, default_flow_style=False)
 
+    def call_salt(self, command):
+        """Call a Salt command."""
+        raise NotImplementedError
```
lorengordon, line 246: I'm not seeing anywhere this is used.

```suggestion
```

src/watchmaker/workers/salt.py
```diff
@@ -774,10 +787,24 @@ def _configuration_validation(self):
                     'parameter `git_repo` was not provided.'
                 )
 
+    def call_salt(self, command):
+        """Call a Salt command."""
+        return self.call_process(command)
```
lorengordon, line 792: Removing because this is not used.

```suggestion
```

src/watchmaker/workers/salt.py
```diff
@@ -928,7 +955,19 @@ def __init__(self, *args, **kwargs):
             'winrepo_dir': os.sep.join((self.salt_win_repo, 'winrepo'))
         }
 
+    def call_salt(self, command):
+        """Call a Salt command."""
+        return self.call_process(command)
```
lorengordon, line 960: Removing because this is not used.

```suggestion
```

src/watchmaker/workers/salt.py
```diff
@@ -928,7 +955,19 @@ def __init__(self, *args, **kwargs):
...
+
     def _install_package(self):
+        if os.path.exists(self.salt_call) and self._check_salt_version():
+            return
+
+        if self._check_salt_version():
+            self.log.info('Salt version %s is already installed.',
+                          self.salt_version)
+            return
```
lorengordon, line 969: This test is handled already by the prior `if` statement.

```suggestion
```

src/watchmaker/workers/salt.py
```diff
@@ -774,10 +787,24 @@ def _configuration_validation(self):
...
```
lorengordon, line 801: This test is handled already by the prior `if` statement.

```suggestion
```

src/watchmaker/workers/salt.py
```diff
@@ -240,6 +241,19 @@ def _prepare_for_install(self):
         ) as fh_:
             yaml.safe_dump(self.salt_conf, fh_, default_flow_style=False)
 
+    def call_salt(self, command):
+        """Call a Salt command."""
+        raise NotImplementedError
+
+    def _check_salt_version(self):
+        current_salt_version = self.call_process(['salt-call', '--version'])
```
lorengordon, line 249: Need to use a value for salt-call that works across both Linux and Windows...

```suggestion
        current_salt_version = self.call_process([self.salt_call, '--version'])
```

lorengordon (changes requested):
Looks like it should _almost_ work for Linux and Windows now. Just one comment on that support.

But there's now a bit of duplicate or unused logic in here to remove.

lorengordon:
go codebuild go

lorengordon:
go codebuild go

lorengordon (approved):

PR #2827, opened 2023-05-02
Checks salt version against the stdout instead of entire call_process return dict

lorengordon:
Example of return structure:

    {'retcode': 0, 'stdout': b'salt-call 3005.1\r\n', 'stderr': b''}


lorengordon:
go codebuild go

lorengordon:
go codebuild go

dwc0011 (approved):

PR #2872, opened 2023-05-09
Add V-244532/RHEL-08-010741 expected-finding caveats

### LINKED ISSUE #2861: [DOCUMENTATION] Some Scanners Report That Files In User Home Directories Are Not Properly Group-Owned
This issue is particularly problematic for enterprise integration tools that typically start up as the `root` user and drop permissions to the service-account's real permission-set. Files created at startup may end up with the `root` user's group-ownership attribute.

Further, the tests that report this finding sometimes note:

image

This finding maps to V-244532 (RHEL-08-010741) and is valid through at least the April 2023 STIG-release([RHEL8v1r10](https://dl.dod.cyber.mil/wp-content/uploads/stigs/zip/U_RHEL_8_V1R10_STIG.zip)).

ferricoxide:
Closes #2861


lorengordon (changes requested):
Just needs to resolve the conflict, then should be good to go

lorengordon (approved):

PR #3124, opened 2023-08-09
Adding ou_path_required config to saltworker

ihenyene:
add an option to the Salt worker, ou_path_required, default it to False. The watchmaker config may set it to True. When True, and when ou_path is not specified (i.e. if ou_path_required and not ou_path), then Watchmaker should exit with an error.


src/watchmaker/workers/salt.py
```diff
@@ -131,7 +131,10 @@ def __init__(self, *args, **kwargs):
         self.salt_debug_log = kwargs.pop('salt_debug_log', None) or ''
         self.salt_content = kwargs.pop('salt_content', None) or ''
         self.salt_content_path = kwargs.pop('salt_content_path', None) or ''
-        self.ou_path = kwargs.pop('ou_path', None) or ''
+        self.ou_path_required = kwargs.pop('ou_path_required', False)
+        self.ou_path = kwargs.pop('ou_path', None) if not self.ou_path_required else kwargs.pop('ou_path', None)
+        if self.ou_path_required and not self.ou_path:
+            raise ValueError("The 'ou_path' option is required, but not provided.")
```
lorengordon, line 137: This test should go in the `before_install()` method

src/watchmaker/workers/salt.py
```diff
@@ -131,7 +131,10 @@ def __init__(self, *args, **kwargs):
         self.salt_debug_log = kwargs.pop('salt_debug_log', None) or ''
         self.salt_content = kwargs.pop('salt_content', None) or ''
         self.salt_content_path = kwargs.pop('salt_content_path', None) or ''
-        self.ou_path = kwargs.pop('ou_path', None) or ''
+        self.ou_path_required = kwargs.pop('ou_path_required', False)
+        self.ou_path = kwargs.pop('ou_path', None) if not self.ou_path_required else kwargs.pop('ou_path', None)
```
lorengordon, line 135: No need to change this
```suggestion
        self.ou_path = kwargs.pop('ou_path', None) or ''
```

lorengordon (commented):
Left a few comments to get things in the right place. Please also add at least one test

lorengordon (approved):

PR #3176, opened 2023-08-29
Update to salt 3006 2

dwc0011:


lorengordon:
@dwc0011 We wouldn't necessarily bump the watchmaker version just for this. We would wait to get tests with the new salt version and the enterprise integrations, which has to be done via the projects that manage the enterprise environments...

src/watchmaker/static/config.yaml
```diff
@@ -1,4 +1,4 @@
-watchmaker_version: ">= 0.27.2.dev"
+watchmaker_version: ">= 0.27.6.dev"
```
lorengordon, line 1: Based on what we know at the moment, this doesn't need a change.

We'd only update the min version here if testing of salt 3006.2 reveals an error that we have to fix in a salt formula, such that the fix would only be packaged in the new version of watchmaker...

lorengordon:
go codebuild go

lorengordon (approved):
Good to go

PR #3182, opened 2023-08-30
Errexit on invalid computer name

dwc0011:


src/watchmaker/__init__.py
```diff
@@ -280,10 +281,25 @@ def __init__(self, arguments):
...
+
+        computer_name = self.config["salt"]["config"].get(
+            "computer_name", None
+        )
+        if computer_name and computer_name_pattern:
+            if not re.match(computer_name_pattern + r"\Z", computer_name):
+                raise WatchmakerError(
+                    "Computer name: %s does not match pattern %s"
+                    % (computer_name, computer_name_pattern)
+                )
```
lorengordon, line 302: In watchmaker, the workers own the "validation" of their arguments. There is a [`before_install()`](https://github.com/plus3it/watchmaker/blob/e777d7b5b57185835c79bcd6b26703a28bdfb112/src/watchmaker/workers/salt.py#L180-L199) we use to perform this kind of check.

tests/test_watchmaker.py
```diff
@@ -19,43 +21,6 @@
...
-@pytest.fixture
-def watchmaker_client(watchmaker_arguments):
-    """Return  watchmaker client with defaults."""
-    return watchmaker.Client(watchmaker_arguments)
-
-
-@pytest.fixture
-def default_config():
-    """Return default configuration for watchmaker."""
-    config_path = str(resources.files(static) / 'config.yaml')
```
lorengordon, line 55: I guess we weren't using this fixture, but the use of the `importlib_resources` library to resolve filepaths was very intentional. I am not sure if it is relevant to the use case here, though. Did you try using the resources interface?

In general, the resources interface is the "correct" way these days in Python to reference paths to static files in a python package, instead of trying to construct them from os.path or Pathlib. The resources interface works for _all_ package types, which the others do not (think zip packages and in-memory runtimes, such as when using pyoxidizer).

dwc0011: I did not use the resources interface, I don't know of a case where I would need to use it.  In my tests there is a tests/resources folder that I do use os.join path on to build the path, but that is a different case than this correct?  I am not importing a package, simply referencing test resource files.

lorengordon: Yeah I just wasn't sure. I don't think the tests are included in the watchmaker package, so maybe we don't have to worry about that particular concern.

dwc0011: Yeah, the package directory is src, so tests are not included, which one would expect.

src/watchmaker/config/__init__.py
```diff
@@ -126,8 +127,27 @@ def get_configs(system, worker_args, config_path=None):
...
+        config.get("salt", {})
+        .get("config", {})
+        .get("computer_name_pattern", {})
+    )
+
+    if computer_name_pattern:
+        try:
+            re.compile(computer_name_pattern)
+        except re.error:
+            raise WatchmakerError(
```
lorengordon, line 151:
```suggestion
            raise InvalidComputerNameError(
```

lorengordon: Oh nevermind. We don't know that the pattern is invalid yet here, just that the regex is not valid.

src/watchmaker/workers/salt.py
```diff
@@ -198,6 +204,14 @@ def before_install(self):
                 "The 'ou_path' option is required, but not provided."
             )
 
+        if self.computer_name and self.computer_name_pattern:
+            if not re.match(self.computer_name_pattern + r"\Z",
+                            self.computer_name):
```
lorengordon, line 209:
```suggestion
            if not re.match(self.computer_name_pattern, self.computer_name):
```

tests/test_watchmaker.py
```diff
@@ -185,45 +129,21 @@ def test_extra_arguments_list(provider_mock):
...
 
     # test
     watchmaker_client = watchmaker.Client(watchmaker_arguments)
 
     # assertions
     assert watchmaker_client.worker_args == check_val
-
-
-@patch("watchmaker.status.Status.initialize", return_value=None)
-def test_none_arguments_aws_provider(provider_mock):
```
lorengordon, line 209: This test actually is doing something... It is validating that the `Arguments()` invocation is converting string values of `"None"` to the pythonic none value `None`...

dwc0011: But it is no different then test_none_arguments test, it does the exact same thing. which is why I removed it.

lorengordon: Ahh, I see

src/watchmaker/workers/salt.py
lorengordon: Do we know it still works correctly if the user specifies a full match pattern themselves? e.g. `(?i)^xyz[\d]{3}[a-z]{8}[ex]$`

dwc0011: Yes, it still works, I have tested it but will add some tests proving it.

lorengordon:
go codebuild go

lorengordon (approved):

PR #3527, opened 2025-09-17
Add Findings FAQ-entry for RHEL-09-611200

ferricoxide:
Add notation to EL9 findings describing scan-errors for rescue-mode config-settings.

### DIFF
--- a/docs/findings/el9.md
+++ b/docs/findings/el9.md
@@ -32,6 +32,7 @@ A few scans performed against EL9 systems are version-dependent. Watchmaker is d
   .. _Enable SSH Server firewalld Firewall Exception: #enable-ssh-server-firewalld-firewall-exception
   .. _Enable Certmap in SSSD: #enable-certmap-in-sssd
   .. _OS library files must have mode 755 or less permissive: #os-library-files-must-have-mode-755-or-less-permissive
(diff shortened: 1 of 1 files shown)


docs/findings/el9.md
```diff
@@ -330,5 +335,31 @@ These are files that _need_ to set to mode `4755` &ndash; permissions that are b
...
+/bin/sh -c "/sbin/sulogin ; /usr/bin/systemctl --fail --no-block default"
+```
+
+Regardless of method chosen, some scanners will still flag the service-configuration as non-compliant. STIG guidance indicate that so long as:
+
+```bash
+grep sulogin /usr/lib/systemd/system/rescue.service
+```
+
+Returns either of the above, two values that the scan-target is adequately configured for this finding.
```
lorengordon, line 362: Just a nit, but should that comma be after "two values"? My brain is hitting a halt and core dump scenario trying to parse as is...

```suggestion
Returns either of the above two values, that the scan-target is adequately configured for this finding.
```

ferricoxide: The "above" and "two" are both modifiers (adjectives) for "values" (noun). Grammar rules state that adjectives in series should be comma-separated (see documents like [this one](https://prowritingaid.com/comma-between-two-adjectives)).

lorengordon (approved):

Issue #3528, opened 2025-09-17
[DOCUMENTATION] Update contribution-guide to remove references to `tox`


ferricoxide:
Use of `tox` has been previously deprecated from the project in favor of other actions implemented via GitHub actions. Need to remove stale references to the `tox` based method for previewing the HTML that was previously generated via `tox`.

PR #3529, opened 2025-09-17
Issue 3528

ferricoxide:
Update process-content

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -39,55 +39,22 @@ To set up `watchmaker` for local development:
 1.  Fork [watchmaker](https://github.com/plus3it/watchmaker) (look for the
     "Fork" button).
 
-2.  Clone your fork locally and update the submodules:
+1.  Clone your fork locally and update the submodules:
 
     ```shell
     git clone https://github.com/your_name_here/watchmaker.git && cd watchmaker
     git submodule update --init --recursive
     ```
 
-3.  Create a branch for local development:
+1.  Create a branch for local development:
 
     ```shell
     git checkout -b name-of-your-bugfix-or-feature
     ```
 
-4.  Now you can make your changes locally.
+1.  Now you can make your changes locally.
 
-5.  When you're done making changes, use [tox][2] to run the linter, the tests,
-    and the doc builder:
-
-    ```shell
-    tox
-    ```
-
-    > **NOTE**: This will test the package in all versions of Python listed in
-    > `tox.ini`. If `tox` cannot find the interpreter for the version, the test
-    > will fail with an `InterpreterNotFound` error. This is ok, as long as at
-    > least one interpreter runs and the tests pass. You can also specify which
-    > [tox environments](#tox-tips) to execute, which can be used to restrict the
-    > Python version required.
-    >
-    > You can also rely on Travis and Appveyor to [run the tests][1] after
-    > opening the pull request. They will be slower though...
-
-6.  In addition to building the package and running the tests, `tox` will build
-    any docs associated with the change. They will be located in the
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -104,40 +74,58 @@ open the pull request.
...
-1.  The _primary_ tox environments for `watchmaker` include:
+## Documentation Previewing
 
-    *   `check`
-    *   `docs`
-    *   `py26`
-    *   `py27`
-    *   `py35`
-    *   `py36`
+In previous iterations of this guide, `tox` was used by the project's CI
```
lorengordon, line 83: I don't think this adds value. Just cover the current process, not the old one.

ci/local/Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+COPY requirements/docs-check.txt requirements/docs-check.txt
+
+RUN dnf install -y python3-pip ; \
+    python3 -m pip install -r requirements/pip.txt ; \
+    python3 -m pip install -r requirements/basics.txt ; \
+    python3 -m pip install -r requirements/docs.txt ; \
+    python3 -m pip install -r requirements/docs-check.txt ; \
+    python3 -m pip install myst_parser ; \
+    python3 -m pip install sphinx-rtd-theme ; \
+    python3 -m pip install --upgrade sphinx
```
lorengordon, line 17: These last three are already covered by the docs.txt file. Specifying them separately like this actually means you'd end up using different versions than pinned by the watchmaker project.

ferricoxide:
> > At any rate, should I go ahead and nuke the (presumably stale/disused) tox directory-content? This stuff:
> > ```
> > ./.tox
> > ./.tox/docs/.tox-info.json
> > ./.tox/.pkg/.tox-info.json
> > ```
> 
> You can do whatever you like with the .tox directory. It's always been in .gitignore, so it's not considered part of the project. It's just a cache on the local system.

Ok. I'm gonna nuke it for tidiness' sake.

ci/local/Dockerfile
ferricoxide: I added those because they _weren't_ getting installed when relying solely on the requirements files.

ferricoxide:
I'll look at using a newer Python RPM-set ...but I might just switch to using one of the Python-specific images that are, I think, based on Alpine. I mean, nothing about the container actually requires an RPM-based distro age.

ci/local/Dockerfile
ferricoxide: I'll look at using a newer Python RPM-set ...but I might just switch to using one of the Python-specific images that are, I think, based on Alpine. I mean, nothing about the container actually requires an RPM-based distro age.

ferricoxide: Oh. Good. Lord. Why is a container built from the python:latest image so. Freaking. Fat???

```
$ podman images
REPOSITORY                                   TAG          IMAGE ID      CREATED         SIZE
localhost/doc-preview                        latest       360b4d419e79  10 seconds ago  436 MB
localhost/doc-review/python3.12-alpine       latest       e05851293725  12 minutes ago  209 MB
localhost/doc-review/python-latest           latest       5077ac676c94  18 minutes ago  1.28 GB
```
By way of explanation:

* `localhost/doc-preview`: is the one build from the Alma Linux image
* `localhost/doc-review/python-latest` is the one built from the "python:latest" image
* `localhost/doc-review/python3.12-alpine`: is the one built from the Python 3.12 "alpine" image

Given the size deltas &mdash; and the fact that the script works on all three &mdash; think I'm going to go for the Alpine-based one.

ferricoxide:
Closes #3528

PR #3655, opened 2026-04-13
Uses pinned uv version to get and install latest python patch version

lorengordon:
This works because uv embeds metadata about the latest python versions in each uv release. So we can use the uv release as a proxy for the latest python version. This simplifies the python install process considerably.


lorengordon (commented):
/build

dwc0011 (approved):

Issue #3676, opened 2026-05-05
Support AL2023 / AMIgen9 without downstream ver == 2023 Jinja workarounds


tyesnyder-p3:
Downstream site-localization content currently requires repeated Jinja special-casing for Amazon Linux 2023 / AMIgen9 systems, e.g.:

{%- if ver == 2023 %}
  {%- set ver = 9 %}
{%- endif %}

This workaround appears in multiple files and would be better handled upstream in Watchmaker so downstream content does not need to duplicate this logic.

Request:
- Add first-class AL2023 / AMIgen9 handling in Watchmaker platform detection
- Allow downstream Salt content to rely on normalized version mapping

Context:
Observed during Watchmaker validation work for AL2023 / AMIgen9 EC2s in a multi-distro customer-environment.

lorengordon:
I feel like there is a bit of an [xy problem](https://xyproblem.info/) here. We have something of a solution proposed, but not a sufficient description of the use case or the problem. Can you elaborate on the problem, in detail?

lorengordon:
Closed with 0.31.0. Amazon Linux 2023 is now supported. Give it a try and check the setup for watchmaker-salt-content as a working example of how to structure the mapping for AL2023.

* https://github.com/plus3it/watchmaker/releases/tag/0.31.0

Issue #3449, opened 2025-03-06
Watchmaker support for Amazon Linux 2 and Amazon Linux 2023


brianhatchl:
Has anyone attempted to add Watchmaker support for AL2 and AL2023 AMIs?

lorengordon:
Hi @brianhatchl, it's been a while since any of us looked seriously at using any of the Amazon Linux distros with Watchmaker... Going waaaay back to the original `amazonlinux`, we did try, but it failed to support FIPS or SELinux, so we set it aside. There are essentially three things we need first, before it makes sense to put the time to it:

* FIPS Support (done in AL2023)
* SELinux Support (done in AL2023)
* An OpenSCAP Profile with remediation support, preferably distributed via [ComplianceAsCode](https://github.com/ComplianceAsCode/content) (Done recently-ish in [v0.1.74](https://github.com/ComplianceAsCode/content/releases/tag/v0.1.74))

So, in theory we could get things going, at this point. Which does highlight a 4th thing, a client or other business development project to cover the time. Or a really motivated user, submitting pull requests!

(Beyond the basic hardening, another roadblock to using AL2023 in an enterprise would likely also require support from all the various security agents used by the enterprise, e.g. splunk, nessus, forescout, etc... But that wouldn't stop Watchmaker from claiming support for the hardening function...)

brianhatchl:
Hi @lorengordon, that sounds like good news.  I anticipate we may end up needing this feature to get ATO for a program that uses AWS Elastic MapReduce (EMR).  The nodes in those clusters only support Amazon Linux flavors of OS.

I have had luck installing OpenSCAP on AL2/AL2023, and it had a profile for amazon linux and I was able to generate a STIG Report (pretty low score in the 60s 😬 out of the box).  I'm not sure if this profile is one with remediation support.  I was also able to install Nessus agent and get a vulnerability scan for it.  Have not tried installing Splunk yet, but 🤞 it may just work.

We have some python developers on the program, so will probably see if we could add AL support and submit a PR.  If we end up stuck or end up with less time than money I'll reach out about funding the effort.

Thanks!

lorengordon:
I can offer some coaching on _where_ to add support... Watchmaker source is a submodule wrapper around a number of salt formulas, so the PR contributions are a little spread out. Check [.gitmodules](https://github.com/plus3it/watchmaker/blob/main/.gitmodules) for complete list, but here's a few that definitely need updating:

* watchmaker-salt-content
[...]

PR #3711, opened 2026-07-13
Releases watchmaker 0.31.0

lorengordon:
Closes:
* https://github.com/plus3it/watchmaker/issues/3676
* https://github.com/plus3it/watchmaker/issues/3449

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -253,9 +253,10 @@ Parameters supported by the Yum Worker:
 
   * `dist` (_list_): Distributions that would install this repo. Some repos
         are supported by multiple distros. (Currently supported distros are
-        redhat, centos, and amazon.)
+      almalinux, centos, oracle, redhat, rocky, and al2023.)
   * `el_version` (`_string_`): The Enterprise Linux version for this repo,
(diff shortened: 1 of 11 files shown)


lorengordon (commented):
/build

docs/configuration.md
```diff
@@ -301,7 +302,8 @@ This example can be used to construct your own `config.yaml` file. The
 versions.
 
 ```yaml
-watchmaker_version: ">= 0.24.0.dev"
+watchmaker_version: ">= 0.27.2.dev"
```
dwc0011: Since this is being updated is there any reason not to make it be 0.30.0.dev 
since I would presume earlier versions do not support al2023?

lorengordon: That's probably a good idea considering the human semantics of it. But the technical semantics of the argument are intended to represent whether there are any inputs of the config.yaml _itself_ that require a particular watchmaker version. There's really no changes to the config that meet that definition...

lorengordon: Updated

dwc0011 (approved):

lorengordon (commented):
/build

Issue #3735, opened 2026-08-21
[DOCUMENTATION] Remove references to EL 7


ferricoxide:
Red Hat ended maintenance-support for 7.X in May of 2024. Similarly, EL7 remediation-content was [removed](https://github.com/plus3it/ash-linux-formula/commit/f77046f4b1ef012ab4dbc765d8cde48bb2514702) from the ash-linux-formula contents in July of 2025.

Need to remove references to EL7 from the watchmaker documentation to ensure people don't continue to assume its availability in watchmaker.

PR #3736, opened 2026-08-21
Documentation/issue 3735

ferricoxide:
Addresses #3735 by removing references to EL distro-versions older than EL8


docs/troubleshooting/Linux/cloud-init.log.md
```diff
@@ -9,7 +9,7 @@
 
 # The `/var/log/cloud-init.log` Log-File
 
-This is the default location where the Red Hat packaged version of the `cloud-init` service for Enterprise Linux 6 and 7 writes _all_ of its log-output to &ndash; on RHEL 8+, logging data is split-out across this file and the `/var/log/cloud-init-output.log` file. All automation directly-initiated through `cloud-init` and that emits STDOUT and/or STDERR messages will be duplicated here. 
+On RHEL 8+, logging data is split-out across this file and the `cloud-init.log` file. All automation directly-initiated through `cloud-init` and that emits STDOUT and/or STDERR messages will be duplicated into the `/var/log/cloud-init.log`.  The `cloud-init.log` file captures the `cloud-init` service's internal state, module execution, and framework logic of cloud-init itself
```
lorengordon: The prior content was referring to "this file and the `/var/log/cloud-init-output.log` file"... 

Can you re-review the intent here? The section heading is "# The `/var/log/cloud-init.log` Log-File"... It's not making sense to say data is split across "this file" (referring to the file in the section), and the file `cloud-init.log` (which is the same file),

ferricoxide: Bleah. Was trying to remove the leftover EL 6/7 logic ...but the previous verbiage already wasn't great. Lemme see if I can clean it up differently.

ferricoxide: Actually, given what it no longer contains, that one can just be omitted from troubleshooting.

lorengordon (dismissed):

lorengordon:
@ferricoxide oops, failure is from the remaining doc ref to the now-deleted cloud-init.log.md file...

https://github.com/plus3it/watchmaker/blob/ff50503b6d7f8ee20d75bbaabbd6e33f0fdc41d0/docs/troubleshooting/Linux/watchmaker.log.md?plain=1#L15

ferricoxide:
> @ferricoxide oops, failure is from the remaining doc ref to the now-deleted cloud-init.log.md file...
> 
> https://github.com/plus3it/watchmaker/blob/ff50503b6d7f8ee20d75bbaabbd6e33f0fdc41d0/docs/troubleshooting/Linux/watchmaker.log.md?plain=1#L15

That'll teach me not to run the local documentation-lint before pushing.

lorengordon (approved):

